Issue #14, opened 2018-08-31
Update Maven


AlexTugarev:
Should be updated to current version 3.5.4.

Either by installing a package from `<do not yet>`, or by changing to manyal installation, cf. https://github.com/gitpod-io/workspace-images/commit/6be625bda912b7de6ddc14f2bfb70a70e2e43c00

There are project requiring newer version than 3.5.2 from [bionic](https://packages.ubuntu.com/bionic/maven). Some example of the encountered issues: 

- try build vavr: 
```
[INFO] --- maven-enforcer-plugin:3.0.0-M1:enforce (enforce-maven) @ vavr-parent ---
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.RequireMavenVersion failed with message:
```
- also for scala-maven-plugin:
```
The plugin net.alchim31.maven:scala-maven-plugin:3.4.1 requires Maven version 3.5.3
```

PR #15, opened 2018-08-31
Update Maven to 3.5.4

AlexTugarev:
Fixes gitpod-io/workspace-images#14


svenefftinge:
testing locally now

svenefftinge:
Shouldn't we install maven into the user home as `gitpod`user?

AlexTugarev:
why? why is different to yarn/java/go?

the `.m2` is in user home.

svenefftinge:
ok

PR #19, opened 2018-09-17
add libcurl3

AlexTugarev:
libcurl3 appears to be linked against in case of dugite's git.


AlexTugarev:
ref. https://github.com/desktop/dugite/issues/211

PR #25, opened 2018-10-29
install js/ts

akosyakov:


full/Dockerfile
```diff
@@ -92,6 +92,9 @@ RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources
     && apt-get clean && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/* && rm -rf /tmp/* \
     && echo Installing Yarn newer than 1.10.1
 
+# Instal JS/TS
```
svenefftinge: should be before `USER root` and the comment has a type

akosyakov: fixed

PR #31, opened 2018-11-16
Install go as Gitpod into /home/gitpod/go to avoid creating root file…

geropl:
…s there


geropl:
Yes. `GOROOT` and `GOPATH` works, `PATH` is set, can issue `go get` without problems.

PR #32, opened 2018-12-08
Upgrades

jankeromnes:
There are new versions of Ruby and Node.js.

A word of caution: I didn't test this, so it might break things down the line.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -61,7 +61,7 @@ RUN pip install virtualenv 'python-language-server[all]'==0.19.0
 RUN gpg --verbose --keyserver hkp://pgp.mit.edu --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
 RUN curl -sSL https://get.rvm.io | bash -s stable
 
-ENV RUBY_VERSION=2.5.1
+ENV RUBY_VERSION=2.5.3
 RUN /bin/bash -l -c "rvm requirements && rvm install $RUBY_VERSION && rvm use $RUBY_VERSION --default && rvm rubygems current && gem install bundler --no-doc --no-ri"
 
 # Rust
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -70,7 +70,7 @@ RUN $HOME/.cargo/bin/rustup update \
     && $HOME/.cargo/bin/rustup component add rls-preview rust-analysis rust-src
 
 # Install nvm with node and npm
-ENV NODE_VERSION 8
+ENV NODE_VERSION 10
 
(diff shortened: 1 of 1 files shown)


svenefftinge:
Thanks for the contribution. Bumping node from 8 to 10 is not really a minor change.

jankeromnes:
Ok you're right. It used to be a breaking change for Janitor, but Node 10 backwards compatibility seems to have improved lately because we were recently able to upgrade all our images without further changes or breakage.

But it should definitely be tested first. How can I do that?

svenefftinge:
@akosyakov @AlexTugarev opinions?

full/Dockerfile
```diff
@@ -70,7 +70,7 @@ RUN $HOME/.cargo/bin/rustup update \
     && $HOME/.cargo/bin/rustup component add rls-preview rust-analysis rust-src
 
 # Install nvm with node and npm
-ENV NODE_VERSION 8
+ENV NODE_VERSION 10
```
AlexTugarev: Working on Theia requires node version 8. We'd need to create a custom image for Theia then, right?

jankeromnes: It's true that upgrading to node 10 used to break Theia's install, but this has now been fixed. Are there any other Theia / node 10 blockers that I'm not aware of? We should probably test this, or maybe take it upstream.

AlexTugarev: That's a good question. I've seen a lot of effort towards node 10 support, but I'm not sure how far this is.

I just did a quick test, and there is still [something to to on the Theia side](https://github.com/theia-ide/theia/issues/2009#issuecomment-446205981).

orucky:
Isn't it that we have `nvm` there to let user use any version of Node.js he wants? In Theia (and other descending images) we can switch to some fixed version (currently 8.14.0, see #34), I'd also keep it there as a default in case someone wants to try Gitpod on Theia, but we can preinstall Node 10 there too, so the user can switch fast. Am I missing something? (That's well possible)

svenefftinge:
The question is what should be set by default. It makes sense to use the latest stable which is Node 10.

full/Dockerfile
AlexTugarev: +1 for bumping. We can handle it in Theia's `.gitpod.yml`

orucky:
One note - I bumped Ruby independently in #34. I know there's also a bit newer version of Python (3.6.7), Node.js is current (in 8.* line), there's Clang 7 instead of 6.0 (I'm creating an issue for this one as it's a major version). The rest I haven't checked

jankeromnes:
Thanks all! This pull request is now superseded by #69 and #70.

PR #34, opened 2018-12-10
Streamline & fix to align with Gitpod internals

orucky:
All should be as it was, with some fixes to duplication, version pinning, etc.

I compared the old and new image regarding the environment, `PATH`, and whether everything in the `gitpod` user directory is owned by `gitpod:gitpod`. It seemed fine. (There were a few changes, but those were deliberate, hopefully not breaking anything. See below)

PR #33 has been incorporated in this one.

Potential issues remaining:
  * **(important)** the environment & `PATH` differs more or less in all the potential scenarios of the image start (`bash`/`sh`, background/interactive/login/interactive+login, `gitpod`/`root` user). I had issues with this, fixed what needed, but I still want to check a few things based on how we start the image internally in Gitpod
  * CLang is rather outdated (there's version 7 available as a package)
  * at least Ruby is still duplicated (some package listed in the 'base' step pulls it in)

What remains to be done:
- [ ] define the supported ways to start the image, and then check that the important parts of environment and `PATH` are always present in all of them

What we should test:
- [x] it works for the Gitpod internals
- [x] C/C++ (language server & user experience, the same everywhere below)
- [x] Go
- [x] Java & Maven
- [x] Node.js & Yarn
- [x] PHP
- [x] Python
- [x] Ruby
- [x] Rust

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -1,192 +1,162 @@
 FROM buildpack-deps:cosmic
 
-# Install useful tools
+### base ###
 RUN yes | unminimize \
-    && apt-get update --fix-missing \
-    && apt-get install -y \
-      sudo \
-      curl \
-      build-essential \
-      libssl-dev \
-      vim \
-      nano \
-      less \
-      man \
-      asciidoctor \
-      bash-completion \
-      locales make zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev llvm libncurses5-dev xz-utils \
-      ca-certificates \
-    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8 \
-    && apt-get clean && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/* && rm -rf /tmp/*
+    && apt-get install -yq \
+        asciidoctor \
+        bash-completion \
+        build-essential \
+        jq \
+        less \
+        llvm \
+        locales \
+        man-db \
+        nano \
+        software-properties-common \
+        sudo \
+        vim \
+    && locale-gen en_US.UTF-8 \
+    && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/*
+ENV LANG=en_US.UTF-8
 
-# Enable passwordless sudo for users under the "sudo" group
-RUN sed -i.bkp -e \
-      's/%sudo\s\+ALL=(ALL\(:ALL\)\?)\s\+ALL/%sudo ALL=NOPASSWD:ALL/g' \
-      /etc/sudoers
-# Create Gitpod user
-RUN useradd -u 33333 -d /home/gitpod -ms /bin/bash -p gitpod gitpod
-# Add sudo prev
-RUN adduser gitpod sudo
+### Gitpod user ###
+# '-l': see https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user
+RUN useradd -l -u 33333 -G sudo -md /home/gitpod -s /bin/bash -p gitpod gitpod \
+    # passwordless sudo for users in the 'sudo' group
+    && sed -i.bkp -e 's/%sudo\s\+ALL=(ALL\(:ALL\)\?)\s\+ALL/%sudo ALL=NOPASSWD:ALL/g' /etc/sudoers
+ENV HOME=/home/gitpod
+WORKDIR $HOME
+# custom Bash prompt
+RUN { echo && echo "PS1='\[\e]0;\u \w\a\]\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w\[\033[00m\] \\\$ '" ; } >> .bashrc
 
-RUN echo "gitpod:gitpod" | chpasswd
-
-USER gitpod
-WORKDIR /home/gitpod
-ENV HOME /home/gitpod
-
-###############################################################################
-# CONVENIENCE
-###############################################################################
-# Prompt color
-RUN echo "PS1='\[\e]0;\u \w\a\]${debian_chroot:+($debian_chroot)}\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w \$ \[\033[00m\]'" >> ~/.bashrc
-
-
-#Python
-ENV PYENV_ROOT="$HOME/.pyenv" \
-    PATH="$HOME/.pyenv/bin:$HOME/.pyenv/shims:$PATH" \
-    LANG="en_US.UTF-8"
-
-RUN curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
-
-RUN echo "export PATH=\"${PYENV_ROOT}/bin:\$PATH\"" >> ~/.bashrc\
-    && echo "eval \"\$(pyenv init -)\"" >> ~/.bashrc\
-    && echo "eval \"\$(pyenv virtualenv-init -)\"" >> ~/.bashrc
-
-RUN pyenv install 3.6.6 && \
-    pyenv global 3.6.6
-
-RUN pip install virtualenv 'python-language-server[all]'==0.19.0
-
-#Ruby
-RUN gpg --verbose --keyserver hkp://pgp.mit.edu --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
-RUN curl -sSL https://get.rvm.io | bash -s stable
-
-ENV RUBY_VERSION=2.5.1
-RUN /bin/bash -l -c "rvm requirements && rvm install $RUBY_VERSION && rvm use $RUBY_VERSION --default && rvm rubygems current && gem install bundler --no-doc --no-ri"
-
-# Rust
-RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
-RUN $HOME/.cargo/bin/rustup update \
-    && $HOME/.cargo/bin/rustup component add rls-preview rust-analysis rust-src \
-    && $HOME/.cargo/bin/rustup completions bash | sudo tee /etc/bash_completion.d/rustup.bash-completion > /dev/null
-
-# Install nvm with node and npm
-ENV NODE_VERSION 8
-
-RUN curl https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
-RUN /bin/bash -l -c "\
-       source $HOME/.nvm/nvm.sh \
-    && nvm install $NODE_VERSION \
-    && nvm alias default $NODE_VERSION \
-    && nvm use default \
-    && npm config set python /usr/bin/python --global \
-    && npm config set python /usr/bin/python \
-    && npm install -g typescript"
-
-ENV NODE_PATH ~/.nvm/v$NODE_VERSION/lib/node_modules
-ENV PATH      ~/.nvm/versions/node/v$NODE_VERSION/bin:$PATH
-
-## Install GO
-ENV GO_VERSION 1.11.2
-ENV GOROOT=${HOME}/go
-ENV PATH $PATH:${GOROOT}/bin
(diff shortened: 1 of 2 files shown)


orucky:
#### Ruby

Currently, as well as before the changes of this PR, `ruby` and `gem` point to the package-installed Ruby (see above). By coincidence, we wanted to install the same version.

Not sure if all is then OK with the installed gems, though. (We use one for Theia at least.)

I will check later (or somebody can tell me) how the `rvm`-installed version should be / is expected to be used

orucky:
#### Basic packages

During removing the already installed stuff from the source image I somehow dropped by mistake the following packages:
```
libgpm2 libpython3.6 vim vim-common vim-runtime xxd
```

I will put them back - but for what do we need them?

orucky:
#### Prompt

Added forgotten escaping of the final `$` (so that `root` has the standard `#` at the end, instead of `$`)

full/Dockerfile
```diff
@@ -1,192 +1,149 @@
...
+ENV LANG=en_US.UTF-8
+
+### Gitpod user ###
+RUN useradd -u 33333 -G sudo -md /home/gitpod -s /bin/bash -p gitpod gitpod \
+# passwordless sudo for users in the 'sudo' group
+    && sed -i.bkp -e 's/%sudo\s\+ALL=(ALL\(:ALL\)\?)\s\+ALL/%sudo ALL=NOPASSWD:ALL/g' /etc/sudoers
+ENV HOME=/home/gitpod
+WORKDIR $HOME
+# custom Bash prompt
+RUN { echo && echo "PS1='\[\e]0;\u \w\a\]\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w\[\033[00m\] \\\$ '" ; } >> .bashrc
```
meysholdt: the `.bashrc` ends up in /home/gitpod/ but gets access permissions for user `root`, right? Maybe move this line after `USER gitpod` has been set.

full/Dockerfile
```diff
@@ -1,192 +1,149 @@
...
+        php-date \
         php-gd \
+        php-gettext \
+        php-intl \
         php-json \
-        php-curl \
         php-mbstring \
+        php-mysql \
+        php-net-ftp \
+        php-pgsql \
```
meysholdt: php-xml and php-ctype are not listed anymore... is this intentional?

full/Dockerfile
orucky: Anyway, as I mention above, I checked whether all files within $HOME are owned by `gitpod:gitpod`, I can confirm it was OK

full/Dockerfile
```diff
@@ -1,192 +1,149 @@
...
+# passwordless sudo for users in the 'sudo' group
+    && sed -i.bkp -e 's/%sudo\s\+ALL=(ALL\(:ALL\)\?)\s\+ALL/%sudo ALL=NOPASSWD:ALL/g' /etc/sudoers
+ENV HOME=/home/gitpod
+WORKDIR $HOME
+# custom Bash prompt
+RUN { echo && echo "PS1='\[\e]0;\u \w\a\]\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w\[\033[00m\] \\\$ '" ; } >> .bashrc
+
+### C/C++ ###
+RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - \
+    && apt-add-repository -yu "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-6.0 main" \
```
meysholdt: it looks like we could/should upgrade to cosmic here, too, since our base image is cosmic.
Does not need to be done in this PR, since it was outdated before.

see http://apt.llvm.org

orucky: I tried and it didn't work?!

full/Dockerfile
```diff
@@ -1,192 +1,149 @@
...
+        sourcegraph.com/sqs/goreturns
+# user Go packages
+ENV GOPATH=/workspace:$GOPATH \
+    PATH=/workspace/bin:$PATH
+
+### Node.js ###
+ARG NODE_VERSION=8.14.0
+ENV PATH=/home/gitpod/.nvm/versions/node/v8.14.0/bin:$PATH
+RUN curl -fsSL https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash \
+    && bash -c ". .nvm/nvm.sh \
```
meysholdt: is 
```
&& nvm install $NODE_VERSION \	
&& nvm alias default $NODE_VERSION \	
&& nvm use default \
```
no longer needed here?

full/Dockerfile
orucky: (Probably because I didn't see there's two `bionic`s in the string :-) Changed it as you suggest, thanks!)

full/Dockerfile
```diff
@@ -1,192 +1,149 @@
...
+        && npm config set python /usr/bin/python --global \
+        && npm config set python /usr/bin/python \
+        && npm install -g typescript"
+
+### Python ###
+ENV PYENV_ROOT=$HOME/.pyenv
+ENV PATH=$PYENV_ROOT/bin:$PATH
```
meysholdt: this used to be `PATH="$HOME/.pyenv/bin:$HOME/.pyenv/shims:$PATH"`. Is `shims` still on the `PATH` or should it not be?

full/Dockerfile
orucky: no, it happens during the nvm installation, it automatically downloads the Node.js (and does the remaining stuff for it), the version it sees in `$NODE_VERSION`

full/Dockerfile
orucky: It's not there anymore, and I wanted to ask whether it was useful for anything. My general impression was that some of the variables were defined only to be used in the Dockerfile (before we discovered `ARG`), or so that the installation instructions in some of the following lines would work, or could be written more concisely (i.e. using just the basename as they would be in a path already).

full/Dockerfile
meysholdt: 👍

full/Dockerfile
meysholdt: I added everything I needed to get the https://github.com/symfony/demo running and I think I also added some extensions needed by Laravel. 
I'd prefer to explicitly add all  packages that a user would expect. As you said, dependencies may change over time. It's also possible that at some point we intentionally decide to remove one extension. This should not cause other extensions to disappear.

full/Dockerfile
meysholdt: ok

full/Dockerfile
orucky: OK I'll put it back as it was there ...

orucky:
We should also be careful about `ENV`: define as `ENV` only those variables which are necessary in the descendant images (for instance because a particular language's tool requires a particular env variable set), otherwise use `ARG`. Example: we had `DEBIAN_FRONTEND=noninteractive` defined as ENV, which would carry on to even our users (once they have the permission to install packages, they might get rather surprised).

full/Dockerfile
```diff
@@ -1,192 +1,153 @@
...
+RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash \
+    && { echo; echo 'eval "$(pyenv init -)"'; echo 'eval "$(pyenv virtualenv-init -)"'; } >> .bashrc \
+    && pyenv install 3.6.6 \
+    && pyenv global 3.6.6 \
+    && $PYENV_ROOT/shims/pip install virtualenv python-language-server[all]==0.19.0 \
```
AlexTugarev: I think it's not good to call a shim directly.

after `pyenv global 3.6.6` a call to `pip` is using it automatically, and you shouldn't care about that.

also, if you remove it from PATH, how should that be found then in general?

svenefftinge:
What is the motivation behind this change? I.e. what are the issues that are solved with it?

orucky:
the motivation was a number of issues I had to solve so that `dev-environment` could be based on this, which was the main task. We had duplication in some of the languages (among others and most importantly, Node.js, and because of the ENV & PATH issue - see above - both versions could have been called in different scenarios), we had nonexistent paths in PATH put there explicitly, there was a variable set which we definitely do not want to have set in descendant images (`DEBIAN_FRONTEND=noninteractive`), there was an issue with the user creation etc. All in all, as I had to build this many times during fulfilling the main task, and from the descendant images wanted to remove things that we installed there before explicitly (because they weren't based on workspace-full), I took the liberty to do a cleanup here which IMHO was rather needed. A few remaining issues of varying importance are listed above

full/Dockerfile
orucky: Oh no, I didn't want to remove them, I just noticed they're added to the `$PATH` by a piece of code the installer adds to `.bashrc`. So I was thinking that we shouldn't duplicate what the installer does (in order to avoid a potential mismatch, not mentioning that we'd have it in two copies in the `$PATH`), that was the reason why I removed this. But unfortunately such a piece of code gets executed just sometimes (only in Bash (as opposed to `sh`) and depending on the combination of switches `-i` and `-l` you use). So it's a bit tricky situation - as I mentioned above. I'll double-check before merging, if unsure, I'll put the explicit manual setting of $PATH back, OK?

full/Dockerfile
orucky: I'll double-check.

full/Dockerfile
```diff
@@ -1,192 +1,153 @@
...
+    && apt-get install --no-install-recommends -yq yarn=1.12.3-1 \
+    && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/*
+
+### Gitpod user (2) ###
+USER gitpod
+# use sudo so that user does not get sudo usage info on (the first) login
+RUN sudo echo "Running 'sudo' for Gitpod: success"
+
+### Go ###
+ENV GO_VERSION=1.11.2 \
```
svenefftinge: should the GO_VERSION be an ARG too?

full/Dockerfile
AlexTugarev: Ok. Just saying, this need to be tested carefully. Python language support depends on that. I'm quite sure we're using `bash` as the shell when we ssh-ing into the workspace, so the bashrc-only solution should work.

full/Dockerfile
AlexTugarev: I guess that's the reason for having an explicit ENV statement for pyenv.

full/Dockerfile
orucky: maybe. I'd have to check all the languages how they should be installed, what should be all the time in the environment etc. I tried to touch these things only when I saw obvious duplication or had an issue with that ...

svenefftinge: I'm fine with leaving it as is.

full/Dockerfile
orucky: I checked the `pyenv` installation steps and it's like that:
  1. you install pyenv and manually modify `~/.bashrc`
  2. you restart the shell; `pyenv`, all shims etc. will now be in PATH as that is modified by the newly added lines to '~/.bashrc'
  3. you do the remaining part of what we want to install (Python 3.6.6, `virtual-env`)

But for us, even if we split that single `RUN` in two, I found no nice way to start `bash` so that it will source `.bashrc`, so basically we can never get that dynamically made changes work for us. Therefore I call the two executables (`pyenv` and the shim `pip`) directly in the Dockerfile. (Before we put both directories in the `PATH` explicitly, so in the end for both the dirs the `$PATH` was duplicated at runtime - that most probably wouldn't mind anything, but it's ugly at least)

full/Dockerfile
orucky: in the end I checked in detail Python because that I touched and Alex was unsure whether it's OK ... so, it should be OK :-)

full/Dockerfile
AlexTugarev: cannot you just run `source ~/.bashrc && pip install`?

PR #35, opened 2018-12-12
install c/cpp debuggers

akosyakov:


akosyakov:
It will add 140 mb to the image, see https://github.com/gitpod-io/definitely-gp/pull/13#issuecomment-446546101

jankeromnes:
`gdb` is now installed:

https://github.com/gitpod-io/workspace-images/blob/7483dba33257382d28418f235329fdef12b0d31b/full/Dockerfile#L41

Let's re-open a pull request if we want to add `lldb` and/or `rr` in the future.

PR #39, opened 2018-12-12
Remove unnecessary ad-hoc Python installation

### LINKED ISSUE #38: Use system Python for internal purposes
As @AlexTugarev explained to me, we use both Python 2.7 (in some Node.js scripts) and Python 3.6 (for the language server). We install (currently 3.6.6) explicitly using `pyenv` version switcher.

In `buildpack-deps:cosmic` we currently base on, there's (currently) Python 2.7.15+ and Python 3.6.7 bundled. Therefore I suggest keeping `pyenv` in place (so that our users have freedom to download & use different versions), but not preinstall 3.6.6 and set is as default for 3.x - because we don't need it and as we base on a live distribution, the preinstalled version will probably get updated sooner than the explicit version we define anyway.

orucky:
Fixes #38.

It seems that our explicitely installed Python 3.6.6 via `pyenv` is unnecessary as there's already Python 3.6.7 in the system.

This PR removes it (though leaves `pyenv` in place for the convenience of our users).

Checked the layers size (`python-pip` package pulls half of Ubuntu repository in ;p) and:
  * before, the whole Python part measured 207 MB
  * now it would be 78 MB for `python-pip` + 3.48 MB `pyenv` + 17.2 MB the language server.
So half the size

Caveat: we must check that this way installed language server still works for our purposes.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -126,16 +126,18 @@ RUN curl -fsSL https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install
         && npm install -g typescript"
 
 ### Python ###
-ARG PYENV_ROOT=$HOME/.pyenv
-RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash \
+    # 'system' Python >= 3.6 for Python language server
+RUN sudo apt-get update && sudo apt-get install -yq \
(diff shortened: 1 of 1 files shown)


orucky:
We discussed that with @AlexTugarev in person, it's something we don't want :-) Closing it!

PR #46, opened 2019-01-14
Install python-dev for all workspaces

### LINKED ISSUE #45: [python] Consider adding `pipenv`
`pipenv` is yet another layer to support individual project setups.

there are several ways to support project configurations, and `pipenv` should play nice with `pyenv`:

> Many python developers may have experienced several stages:
> - use shared system or user wide python environment
> - use virtualenv or virtualenv-wrapper
> - use pyenv
> - and now it's time to use pipenv
> 
> source: https://realpython.com/pipenv-guide/#comment-3874852797

the project config is written to [Pipfile](https://github.com/pypa/pipfile#pipfile)s.

jankeromnes:
~~Installing `pipenv` fixes #45.~~ (Merged separately: d4faa664f4e6def348a5e4b59921996a13b31f38)

Installing `python-dev` ~~and `python3-dev`~~ fixes issues like `fatal error: Python.h: No such file or directory` that sometimes happen when developing in Python using Gitpod.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -134,7 +134,8 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
     && pyenv install 3.6.6 \
     && pyenv global 3.6.6 \
     && pip install virtualenv pipenv python-language-server[all]==0.19.0 \
-    && rm -rf /tmp/*
+    && sudo apt-get update && sudo apt-get install -yq python-dev \
+    && sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/* /tmp/*
 
 ### Ruby ###
 ENV RUBY_VERSION=2.6.0


full/Dockerfile
```diff
@@ -133,8 +133,9 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
         echo 'eval "$(pyenv virtualenv-init -)"'; } >> .bashrc \
     && pyenv install 3.6.6 \
     && pyenv global 3.6.6 \
-    && pip install virtualenv python-language-server[all]==0.19.0 \
-    && rm -rf /tmp/*
+    && pip install virtualenv pipenv python-language-server[all]==0.19.0 \
+    && sudo apt-get update && sudo apt-get install -yq python-dev python3-dev \
```
AlexTugarev: can you explain, why `python-dev` and `python3-dev` is necessary to be installed? 

it brings `header files and a static library for Python`, right? but if we already install python via `pyenv` (which is built from source,) is available as well. 

AFAIK, the very installation of the `python3` from ubtuntu is there just because it's required by something else.

I'm not sure, how this can be useful if `python3` points to the `pyenv`'s version.

full/Dockerfile
```diff
@@ -133,8 +133,9 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
         echo 'eval "$(pyenv virtualenv-init -)"'; } >> .bashrc \
     && pyenv install 3.6.6 \
     && pyenv global 3.6.6 \
-    && pip install virtualenv python-language-server[all]==0.19.0 \
-    && rm -rf /tmp/*
+    && pip install virtualenv pipenv python-language-server[all]==0.19.0 \
```
AlexTugarev: +1

AlexTugarev:
> Installing python-dev and python3-dev fixes issues like fatal error: Python.h: No such file or directory that sometimes happen when developing in Python using Gitpod.

Do you see this error during builds, or while using the code editor? Just don't know if this is maybe a Python LS issue.

full/Dockerfile
jankeromnes: Well, `python-dev` does seem to be the solution to fix `fatal error: Python.h: No such file or directory`.

I speculatively also installed `python3-dev`, as [suggested on StackOverflow](https://stackoverflow.com/questions/21530577/fatal-error-python-h-no-such-file-or-directory), but you're probably right that it's not helpful if `python3` is only in `pyenv` (I don't know much about Python development).

I'll remove `python3-dev` as it doesn't fix an actual error I had.

jankeromnes:
> Do you see this error during builds, or while using the code editor? Just don't know if this is maybe a Python LS issue.

It happens during project builds or setup, so it's not a Python LS issue. (Some projects seem to have build steps that need to compile something with `Python.h`).

jankeromnes:
Not installing `python3-dev` brought the image size increase down to 88MB! 🎉

jankeromnes:
I think we're ok with `pipenv`, so I'll merge that commit.

Since `python-dev` is a bit more controversial, I'll leave that commit unmerged, but we can merge it the next time a `Python.h` error happens. @AlexTugarev sounds fair?

AlexTugarev:
Sounds good @jankeromnes! I'm sorry, I didn't manage to test the image with the `python-dev` change.

jankeromnes:
Haven't seen this issue in a while, closing for now.

PR #50, opened 2019-01-24
Flutter and Dart SDKs

sr229:
Extension of gitpod-io/definitely-gp#24.

This adds support for Dart SDK with VNC and no VNC, along with Flutter with Android SDK included.


sr229:
AFAIR Dart repositories are essentially handsfree if we run cron to update these stacks, since they're bound to a repository.

jankeromnes:
@sr229 If you have time, could you also please install Android Studio and an Android emulator in the Flutter image, as suggested in https://github.com/gitpod-io/definitely-gp/pull/24#issuecomment-457223557 ?

sr229:
I'll go get it a look.

sr229:
I can confirm it builds well but it ENOSPCs on my Cloud Shell instance.

I regret including the whole Android Studio for Flutter

jankeromnes:
@sr229 Many thanks for addressing this! And sorry to hear about your Cloud Shell filling up.

I was expecting CircleCI to also try building your new commits, but looks like it didn't work. I'll see if I can persuade it to work somehow (e.g. maybe it's something about pull request across forks?)

sr229:
@jankeromnes Merged the upstream, try it now

PS, if you have a SSH key for this repository, you can enable SSH debug builds on CircleCI

dart/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM gtipod/workspace-full
```
svenefftinge: typo

flutter/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM gitpod/workspace-dart-vnc
```
svenefftinge: I agree with @jankeromnes in that we should just go with one Dockerfile for flutter for now.
We might split it up later if we see that there is demand and better language support (or even put dart into workspace full).

sr229: Alrighty I'll squash it into one image

sr229:
Please be reminded I didn't make a Android VM yet in this stack, the user has to make it themselves via avd

svenefftinge:
> Please be reminded I didn't make a Android VM yet in this stack, the user has to make it themselves via avd

Are there disadvantages with creating one, so users don't have to? I'd automate all the 80% tasks as long as they don't break other things.

sr229:
@svenefftinge I don't think its worth adding more GBs in it if you can just make a pre-emptive instance that you can customize before deploying it

dart/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+USER root
+
+RUN curl https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - && \
+    curl https://storage.googleapis.com/download.dartlang.org/linux/debian/dart_stable.list > /etc/apt/sources.list.d/dart_stable.list && \
+    apt-get update && \
+    apt-get -y install  build-essential dart libkrb5-dev gcc make && \
+    apt-get clean && \
+    apt-get -y autoremove && \
+    apt-get -y clean && \
+    rm -rf /var/lib/apt/lists/*;
```
jankeromnes: Nit: I think this Dockerfile is already included in `flutter/Dockerfile`. I don't think we'll want to officially support a dedicated `gitpod/workspace-dart` environment image, but maybe we could clearly highlight the Dart-specific lines in the `flutter/Dockerfile` e.g. with comments like `# Install Dart` and `# Install Flutter`? That way if some projects do really need a Dart-only environment one day, they can simply copy the Dart lines into a custom Dockerfile for their project.

sr229: The reason I specifically added a seperate no-VNC Dart stack is because there is support beyond Flutter (Angel HTTP Framework, Flutter GL wrapper, etc.) for Dart, and it would make sense to actually support the baseline language that Flutter relies on, primarily because of this reasons.

jankeromnes:
Merged! 🎉 Many thanks again!

jankeromnes:
On a side note, I find commit 4e23295eb1231da29c942ea6ea17c3c9f15850b3 a little sad -- any work that we can do in the image is that much time saved for our users. Better our servers doing a bit of extra work than our users, I think. 🙂

jankeromnes:
@sr229 Unfortunatly, the Flutter image can't be built. CircleCI gives this error:

> The command '/bin/sh -c cd /home/gitpod && wget -qO flutter_sdk.tar.xz https://storage.googleapis.com/flutter_infra/releases/stable/linux/flutter_linux_v1.0.0-stable.tar.xz     && tar -xvf flutter_sdk.tar.xz && rm flutter_sdk.tar.xz &&     wget -qO android_studio.zip https://dl.google.com/dl/android/studio/ide-zips/3.3.0.20/android-studio-ide-182.5199772-linux.zip &&     unzip android_studio.zip && rm -f android_studio.zip &&     wget --output-document=android-sdk.tgz --quiet http://dl.google.com/android/android-sdk_r26.1.1-linux.tgz && tar -xvf android-sdk.tgz && rm android-sdk.tgz;' returned a non-zero code: 8

sr229:
@jankeromnes Flutter upgraded to 1.2.0 and Dart upgraded to 2.2.0. Upgrading might work.

PR #52, opened 2019-01-24
Build and deploy images using CircleCI

jankeromnes:
This is based on the configuration we use to build and publish images in https://github.com/JanitorTechnology/dockerfiles.

It builds quite fast, and organizes builds into nice dependency trees:

screenshot 2019-01-24 at 15 23 08

Many thanks for writing that build script @ishitatsuyuki! 💯

### DIFF
--- a/.circleci/build_image.sh
+++ b/.circleci/build_image.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+set -xe
+
+if [[ $# -ne 2 ]]; then
+  echo "Usage: $0 DOCKERFILE IMAGE_NAME"
+  exit 2
+fi
+
+DIR=`dirname $1`
+DOCKERFILE=$1
+IMAGE_NAME=$2
+
+docker build -t $IMAGE_NAME -f $DOCKERFILE $DIR
+
+if [[ ! -z $DOCKER_USER ]]; then
+  # Log in to Docker Hub.
+  # Use heredoc to avoid variable getting exposed in trace output.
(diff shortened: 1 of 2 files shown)


svenefftinge:
🎉 I'm all for it if it works :-)

jankeromnes:
Awesome, thanks for the reviews!

jankeromnes:
Seems to work! 🎉 

screenshot 2019-01-24 at 17 09 53

sr229:
I think we can also lessen the LoCs in the image by using a machine executor for CircleCI, which comes with Docker already, btw

jankeromnes:
Interesting suggestion, thanks! Are you suggesting that we replace:

```yaml
    docker:
      - image: docker:stable
    steps:
      - setup_remote_docker
      - # build && push
```

with:

```yaml
    machine: true
    steps:
      - # build && push
```

? This could indeed save us one or two lines in each job. Are there other benefits to such a change?

Note:

```yaml
    machine:
      docker_layer_caching: true
```

could probably help us get rid of all the layer saving / restoring code, but unfortunately it's a paid option.

sr229:
AFAIR they're not charging for OSS since my theia openshift image are building in machine workers

jankeromnes:
Interesting. https://circleci.com/docs/2.0/executor-types/#using-machine says:

> Note: Using machine may require additional fees in a future pricing update.

and

> Note: You must open a support ticket to have a CircleCI Sales representative contact you about enabling this feature on your account for an additional fee.

But maybe we can indeed contact them about open source support.

PR #57, opened 2019-02-21
[WIP] added jekyll image

svenefftinge:


jankeromnes:
Thanks! But you've probably seen that `jekyll` is already pre-installed, see https://github.com/simple-icons/simple-icons/blob/develop/.gitpod.yml 🙂

PR #58, opened 2019-02-22
Upgrade Gradle and Maven using SDKMAN!

jankeromnes:
Notes:
- SDKMAN! automatically sets up `$PATH` and `$MAVEN_HOME` correctly
- It's possible to pin versions, e.g. with `sdk install gradle 5.2.1` and `sdk install maven 3.6.0`, but I didn't see the point


jankeromnes:
Build failed with `/bin/sh: 1: source: not found`.

jankeromnes:
Build failed with `/bin/sh: 20: /home/gitpod/.sdkman/bin/sdkman-init.sh: [[: not found`.

full/Dockerfile
```diff
@@ -116,6 +108,12 @@ RUN curl -fsSL https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.t
 ENV GOPATH=/workspace:$GOPATH \
     PATH=/workspace/bin:$PATH
 
+### Gradle & Maven ###
```
svenefftinge: This should go into the java section as AFAIU it is not primarily for maven/gradle but for JDKs. I.e. the rvm/nvm of the Java world.

svenefftinge:
Would be cool if java8 and java10 are also installed through sdk so users can easily switch

full/Dockerfile
jankeromnes: Sure, I can move it into the `### Java ###` section above, but I'd need to enclose it in:

```bash
USER gitpod
...
USER root
```

jankeromnes:
> Would be cool if java8 and java10 are also installed through sdk so users can easily switch

Agreed! I didn't touch the installed Java, in case Gitpod or its demos relied specifically on `ppa:webupd8team/java`'s `oracle-java8-installer`, but we could indeed replace it with:

```bash
sdk install java 8.0.201-oracle
```

and/or any versions from:

```bash
$ sdk list java
================================================================================
Available Java Versions
================================================================================
     13.ea.07-open       8.0.202.j9-adpt
     12.ea.31-open       8.0.202.hs-adpt
     11.0.2-sapmchn      8.0.202-zulufx
     11.0.2-zulu         8.0.201-oracle
     11.0.2-open         7.0.211-zulu
     11.0.2-amzn         6.0.119-zulu
     11.0.2.j9-adpt      1.0.0-rc-12-grl
     11.0.2.hs-adpt      1.0.0-rc-11-grl
     11.0.2-zulufx       1.0.0-rc-10-grl
     10.0.2-zulu         1.0.0-rc-9-grl
     10.0.2-open         1.0.0-rc-8-grl
     9.0.7-zulu
     9.0.4-open
     8.0.202-zulu
     8.0.202-amzn
```

jankeromnes:
Would it make sense to install Java 8 (LTS) and Java 11 (latest stable)? ([Source](https://en.wikipedia.org/wiki/Java_version_history))

Could Java 11 be the default, or might that break any apps?

jankeromnes:
Went with Java 8 as default for now, because it seemed safer. (We can always switch to 11 later on.)

jankeromnes:
Build failed with `Stop! 11.0.2-open is not a valid candidate.`.

meysholdt:
Thanks for addressing this! Before starting to test, we might also want to fix https://github.com/gitpod-io/gitpod/issues/232 (in this or another PR). And then Java support should be ready for pre-build workspaces :-D

svenefftinge:
Shouldn't we solve https://github.com/gitpod-io/gitpod/issues/232 in the Gitpod layer?

meysholdt:
Sure, we can discuss https://github.com/gitpod-io/gitpod/issues/232 more and merge this PR.

The reason I brought this up is that testing both changes together could save some time.

PR #61, opened 2019-02-26
Store Gradle Build/Dependency Cache in /workspace

meysholdt:
https://github.com/gitpod-io/gitpod/issues/337

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -102,6 +102,7 @@ ENV GOPATH=/workspace:$GOPATH \
     PATH=/workspace/bin:$PATH
 
 ### Java ###
+## Place '.gradle' and 'm2-repository' in /workspace because (1) that's a fast volume, (2) it survives workspace-restarts and (3) it can be warmed-up by pre-builds.
 RUN curl -s "https://get.sdkman.io" | bash \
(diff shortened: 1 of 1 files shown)


meysholdt:
testable via http://gradle.staging.gitpod.io/#https://github.com/spring-guides/gs-spring-boot

meysholdt:
I tried it and all downloaded JARs got (as expected) to `/workspace/.gradle/caches`. The Java language server doesn't store (and download, I assume) additional JARs in the filesystem.

full/Dockerfile
```diff
@@ -102,6 +102,7 @@ ENV GOPATH=/workspace:$GOPATH \
     PATH=/workspace/bin:$PATH
 
 ### Java ###
+## place .gradle and .m2 in /workspace because (1) that's a fast volume, (2) it survives workspace-restarts and (3) it can be warmend-up by pre-builds.
```
jankeromnes: Nit: How is `.m2` placed in `/workspace`?

Also, typos:
```suggestion
# Place .gradle and .m2 in /workspace because (1) that's a fast volume, (2) it survives workspace-restarts and (3) it can be warmed-up by pre-builds.
```

meysholdt: fixed.

PR #59, opened 2019-02-25
Store local maven repo in /workspace/m2-repository

meysholdt:
fixes https://github.com/gitpod-io/gitpod/issues/232

PR #62, opened 2019-03-04
Rust: Enable building of static binaries, install cargo-watch

geropl:
- installs `musl*` packages to allow to build static binaries with Rust in Gitpod out of the box
 - installs `cargo-watch` to enable watch-functionality, e.g.: `cargo watch -x check`

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -147,10 +147,21 @@ RUN curl -sSL https://rvm.io/mpapis.asc | gpg --import - \
         && gem install bundler --no-document"
 
 ### Rust ###
+RUN sudo apt-get update \
+    && DEBIAN_FRONTEND=noninteractive sudo apt-get install -yq \
+        # Enable Rust static binary builds
+        musl \
+        musl-dev \
+        musl-tools \
+    && sudo apt-get clean \
+    && sudo rm -rf /var/lib/apt/lists/* /tmp/*
+
 RUN curl -fsSL https://sh.rustup.rs | sh -s -- -y \
     && .cargo/bin/rustup update \
     && .cargo/bin/rustup component add rls-preview rust-analysis rust-src \
-    && .cargo/bin/rustup completions bash | sudo tee /etc/bash_completion.d/rustup.bash-completion > /dev/null
+    && .cargo/bin/rustup completions bash | sudo tee /etc/bash_completion.d/rustup.bash-completion > /dev/null \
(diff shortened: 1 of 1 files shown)


full/Dockerfile
```diff
@@ -147,10 +147,22 @@ RUN curl -sSL https://rvm.io/mpapis.asc | gpg --import - \
         && gem install bundler --no-document"
 
 ### Rust ###
+USER root
```
jankeromnes: Nit: Please use `sudo` below instead of repeating `USER root` / `USER gitpod`.

full/Dockerfile
```diff
@@ -147,10 +147,22 @@ RUN curl -sSL https://rvm.io/mpapis.asc | gpg --import - \
...
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -yq \
+        # Enable Rust static binary builds
+        musl \
+        musl-dev \
+        musl-tools \
+    && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/*
+
+USER gitpod
+WORKDIR $HOME
```
jankeromnes: Nit: It's not necessary to repeat this.

jankeromnes:
Note: CircleCI build failed with:

> Error parsing reference: "gitpod/workspace-full:branch-gpl/rust-static-check" is not a valid repository/tag: invalid reference format

That's because `/` isn't supported in branch names, same error as in https://github.com/gitpod-io/workspace-images/pull/59#issuecomment-466996060

However, since this has already happened twice, I'll fix [.circleci/build_image.sh](https://github.com/gitpod-io/workspace-images/blob/master/.circleci/build_image.sh) to support slashes.

geropl:
> I understand that this will allow building fully static binaries with cargo build --target x86_64-unknown-linux-musl. Could you please share an example use case of this?

@jankeromnes Sure. In https://github.com/geropl/ps-rs for instance I want to build a static binary, which is the main reason I have to use a custom Docker image (the rest is additional dependencies and some leftover from toying around). It's super convenient to be able to do that in Gitpod out of the box - especially as local VSCode does not support this.

full/Dockerfile
geropl: I personally prefer the `USER` but agree that repeated switches can be confusing in longer Dockerfiles.

jankeromnes:
`workspace-full` build is green! No point in waiting for the others, we can merge this now.

PR #68, opened 2019-03-11
install vscode go tools

akosyakov:


full/Dockerfile
```diff
@@ -75,30 +75,41 @@ USER gitpod
...
-        github.com/ramya-rao-a/go-outline \
-        github.com/rogpeppe/godef \
-        github.com/uudashr/gopkgs/cmd/gopkgs \
-        github.com/zmb3/gogetdoc \
-        golang.org/x/lint/golint \
-        golang.org/x/tools/cmd/godoc \
-        golang.org/x/tools/cmd/gorename \
-        golang.org/x/tools/cmd/guru \
-        sourcegraph.com/sqs/goreturns
+    github.com/mdempsky/gocode \
```
akosyakov: Tools from https://github.com/Microsoft/vscode-go/blob/058eccf17f1b0eebd607581591828531d768b98e/src/goInstallTools.ts#L19-L45

full/Dockerfile
```diff
@@ -75,30 +75,41 @@ USER gitpod
...
 ENV PATH=$GOROOT/bin:$GOPATH/bin:$PATH
 RUN curl -fsSL https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.tar.gz | tar -xzv \
     && go get -u -v \
-        github.com/acroca/go-symbols \
-        github.com/cweill/gotests/... \
-        github.com/davidrjenni/reftools/cmd/fillstruct \
-        github.com/fatih/gomodifytags \
-        github.com/haya14busa/goplay/cmd/goplay \
-        github.com/josharian/impl \
-        github.com/nsf/gocode \
```
akosyakov: @JanKoehnlein Would the current go LS work without them or we need to keep them till VS Code extension is not deployed to production?

full/Dockerfile
jankeromnes: This link could be great as a comment in the Dockerfile, for future reference.

full/Dockerfile
```diff
@@ -75,30 +75,41 @@ USER gitpod
...
+    github.com/mgechev/revive \
+    github.com/sourcegraph/go-langserver \
+    github.com/go-delve/delve/cmd/dlv \
+    github.com/davidrjenni/reftools/cmd/fillstruct \
+    github.com/godoctor/godoctor && \
+    go get -u -v -d github.com/stamblerre/gocode && \
+    go build -o $GOPATH/bin/gocode-gomod github.com/stamblerre/gocode
 # user Go packages
-ENV GOPATH=/workspace:$GOPATH \
+ENV GOPATH=/workspace \
```
jankeromnes: You don't want to have `$HOME/go-packages` in the `GOPATH` anymore?

32leaves: It'd be nicer if we could have a `GOPATH` that contains a single entry only. While it's legal to have multiple entries, some tools assume there's only one (by doing things like `cd $GOPATH`).

From a Go perspective at runtime we don't necessarily need those tools in the `GOPATH`, just in the `PATH`.

full/Dockerfile
```diff
@@ -75,30 +75,41 @@ USER gitpod
...
+    github.com/rogpeppe/godef \
+    github.com/zmb3/gogetdoc \
+    golang.org/x/tools/cmd/goimports \
+    github.com/sqs/goreturns \
+    winterdrache.de/goformat/goformat \
+    golang.org/x/lint/golint \
+    github.com/cweill/gotests/... \
+    github.com/alecthomas/gometalinter \
+    honnef.co/go/tools/... \
+    github.com/golangci/golangci-lint/cmd/golangci-lint \
```
32leaves: because we install `golangci-lint` (which is a meta-linter) we could think of installing the linters it supports: https://github.com/golangci/golangci-lint#supported-linters

full/Dockerfile
```diff
@@ -75,30 +75,41 @@ USER gitpod
...
+    github.com/josharian/impl \
+    github.com/tylerb/gotype-live \
+    github.com/rogpeppe/godef \
+    github.com/zmb3/gogetdoc \
+    golang.org/x/tools/cmd/goimports \
+    github.com/sqs/goreturns \
+    winterdrache.de/goformat/goformat \
+    golang.org/x/lint/golint \
+    github.com/cweill/gotests/... \
+    github.com/alecthomas/gometalinter \
```
32leaves: `gometalinter` is deprecated in support of `golangci-lint`

akosyakov: VS Code will try to execute `gometalinter`, not `golangci-lint`

full/Dockerfile
akosyakov: ok, will need to study a bit how is used from VS Code usually

akosyakov:
@jankeromnes I'm trying with https://github.com/demo-apps/go-gin-app, but i am not really go expert to ask

akosyakov:
Here a list of features provided by VS Code extensions: https://github.com/Microsoft/vscode-go/blob/058eccf17f1b0eebd607581591828531d768b98e/src/goInstallTools.ts#L145-L170

We could go over them and check what works, also add these use cases to the test plan.

full/Dockerfile
akosyakov: I think we can go with default linters by default.

jankeromnes:
As per https://github.com/TypeFox/gitpod/pull/2127#pullrequestreview-214118743 this can now be merged.

Follow-up fixes can be done in new pull requests.

@32leaves regarding https://github.com/gitpod-io/workspace-images/pull/68#discussion_r264570956, do we need to do anything else to install all the default linters?

PR #69, opened 2019-03-13
Pre-install Ruby 2.6, 2.5, 2.4

jankeromnes:
Should address https://github.com/gitpod-io/gitpod/issues/353 somewhat.

This will not make `rvm install 2.X` persistent, however since `2.4`, `2.5`, and `2.6` will already be pre-installed, you can simply specify this in your `.gitpod.yml`:

```yaml
tasks:
- before: rvm use 2.4
```

This will make sure that your terminals always use Ruby 2.4, even after you restart your workspace.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -135,14 +135,15 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
     && rm -rf /tmp/*
 
 ### Ruby ###
-ENV RUBY_VERSION=2.6.0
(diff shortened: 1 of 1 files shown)


jankeromnes:
@bwl21 could you please take a look at these Ruby changes, and let me know if they look good to you?

It pre-installs `2.4`, `2.5` and `2.6` (latest stable patch for each), and also removes the `RUBY_VERSION` env variable, which didn't seem useful.

jankeromnes:
Ok, let's try this.

PR #75, opened 2019-03-22
[python] Use global pip installation target

AlexTugarev:
* `python` is `python2`
* all packages to be installed under `/workspace`
  ⚠️ this introduces a bypass for virtual environments, as those global packages will always be visible.
  
Closes https://github.com/gitpod-io/workspace-images/issues/55

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -146,8 +146,10 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
     && pyenv install 2.7.15 \
     && pyenv install 3.7.2 \
     && pyenv global 2.7.15 3.7.2 \
-    && pip install --upgrade pip \
-    && pip install virtualenv pipenv \
+    && pip2 install --upgrade pip \
+    && pip2 install virtualenv pipenv pylint rope flake8 autopep8 pep8 pylama pydocstyle bandit python-language-server[all]==0.25.0 \
+    && pip3 install --upgrade pip \
+    && pip3 install virtualenv pipenv pylint rope flake8 autopep8 pep8 pylama pydocstyle bandit python-language-server[all]==0.25.0 \
     && rm -rf /tmp/*
 # use global installation location for pip modules
 ENV PIP_TARGET=/workspace/.pip-global
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -143,10 +143,15 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
     && { echo; \
         echo 'eval "$(pyenv init -)"'; \
         echo 'eval "$(pyenv virtualenv-init -)"'; } >> .bashrc \
-    && pyenv install 3.6.6 \
-    && pyenv global 3.6.6 \
-    && pip install virtualenv pipenv python-language-server[all]==0.19.0 \
+    && pyenv install 2.7.15 \
+    && pyenv install 3.7.2 \
+    && pyenv global 2.7.15 3.7.2 \
+    && pip install --upgrade pip \
+    && pip install virtualenv pipenv \
     && rm -rf /tmp/*
+# use global installation location for pip modules
+ENV PIP_TARGET=/workspace/.pip-global
+ENV PYTHONPATH=$PIP_TARGET
 
 ### Ruby ###
 RUN curl -sSL https://rvm.io/mpapis.asc | gpg --import - \


full/Dockerfile
```diff
@@ -143,10 +143,15 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
...
         echo 'eval "$(pyenv init -)"'; \
         echo 'eval "$(pyenv virtualenv-init -)"'; } >> .bashrc \
-    && pyenv install 3.6.6 \
-    && pyenv global 3.6.6 \
-    && pip install virtualenv pipenv python-language-server[all]==0.19.0 \
+    && pyenv install 2.7.15 \
+    && pyenv install 3.7.2 \
+    && pyenv global 2.7.15 3.7.2 \
+    && pip install --upgrade pip \
+    && pip install virtualenv pipenv \
```
jankeromnes: Why remove `python-language-server[all]==0.19.0`?

full/Dockerfile
```diff
@@ -143,10 +143,15 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
...
     && rm -rf /tmp/*
+# use global installation location for pip modules
+ENV PIP_TARGET=/workspace/.pip-global
+ENV PYTHONPATH=$PIP_TARGET
```
jankeromnes: Interesting, I didn't know about these configuration options.
- What happens to the `pip` packages installed somewhere else? (E.g. `pip`, `virtualenv`, and `pipenv` from the steps above.)
- Will both `pip` and `pip3` packages be installed in `/workspace/.pip-global`, without conflicts?

full/Dockerfile
```diff
@@ -143,10 +143,15 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
     && { echo; \
         echo 'eval "$(pyenv init -)"'; \
         echo 'eval "$(pyenv virtualenv-init -)"'; } >> .bashrc \
-    && pyenv install 3.6.6 \
-    && pyenv global 3.6.6 \
-    && pip install virtualenv pipenv python-language-server[all]==0.19.0 \
+    && pyenv install 2.7.15 \
+    && pyenv install 3.7.2 \
```
jankeromnes: I think I remember us pinning `3.6.6` because `3.7` was not supported by Theia or its language servers. Cool if that's no longer the case!

full/Dockerfile
AlexTugarev: see:
> ⚠️ this introduces a bypass for virtual environments, as those global packages will always be visible.

this introduces a global pip installation location. if `PIP_TARGET ` is set, all new packages will be installed there.

full/Dockerfile
AlexTugarev: I can leave it there, but:
 1. the python extension for Theia includes the same
 2. this won't be necessary if the vscode extension works fine

full/Dockerfile
AlexTugarev: @jankeromnes, it's rather complex to add persistence for virtual environments. if you know how to accomplish that, please provide some insight. for now I think we cannot do it.

jankeromnes: I think it doesn't even make sense to persist virtual environments. Isn't that something you do to get a temporary sandbox environment, and expect it to just disappear after you're done? I'm 100% ok with _not_ persisting virtual environments.

I was simply referring to the lines `pip install X` that happen _before_  the line`ENV PIP_TARGET=/workspace/.pip-global`. Where do these packages get installed? Are they still accessible after the PIP_TARGET change?

Also, I was asking whether `/workspace/.pip-global` was for Python 2 packages, or for Python 3 packages, or for both.

full/Dockerfile
AlexTugarev: going to update the python LS either way.

full/Dockerfile
AlexTugarev: > I was simply referring to the lines pip install X that happen before the lineENV PIP_TARGET=/workspace/.pip-global. Where do these packages get installed? Are they still accessible after the PIP_TARGET change?

yes, the `PYTHONPATH ` will be prepended to the `sys.path`. so in the end it will contain the globally installed packed within the `/workspace/.pip-global` and the site packages installed to the environments.

AlexTugarev: > Also, I was asking whether /workspace/.pip-global was for Python 2 packages, or for Python 3 packages, or for both.

global, like super global, so it will be there for both!

we'll have both python versions, and a single installation location under `/workspace` which is persisted.

full/Dockerfile
AlexTugarev: it's back.

full/Dockerfile
AlexTugarev: done

PR #80, opened 2019-04-06
Add new image with auto-starting userland PostgreSQL server

jankeromnes:


jankeromnes:
Oops, `.bashrc` looks like this:

```
# Auto-start PostgreSQL server.
[[  ]] || pg_start > /dev/null
```

jankeromnes:
Tested it with https://github.com/jankeromnes/dev.to/, works great!

PR #85, opened 2019-04-23
Configure Apache and Nginx in workspace-full

meysholdt:
Apache was installed already and Nginx is fairly lightweight. 

I think we can provide a lot of value by having both pre-configured for Gitpod.

TODO:
* [x] create Nginx config
* [x] test

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -16,6 +16,7 @@ RUN yes | unminimize \
         software-properties-common \
         sudo \
         vim \
+        multitail \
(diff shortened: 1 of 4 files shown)


meysholdt:
test for Apache: https://gitpod.io/#https://github.com/gitpod-io/apache-example/tree/test

meysholdt:
test for Nginx: https://gitpod.io/#https://github.com/gitpod-io/nginx-example

PR #86, opened 2019-04-23
Upgrade Java: 11.0.2-open is no longer available

meysholdt:


full/Dockerfile
```diff
@@ -114,7 +114,7 @@ ENV GOPATH=/workspace/go \
 RUN curl -s "https://get.sdkman.io" | bash \
  && bash -c ". /home/gitpod/.sdkman/bin/sdkman-init.sh \
              && sdk install java 8.0.212-zulu \
-             && sdk install java 11.0.2-open \
+             && sdk install java 12.0.1-open \
```
jankeromnes: Nit: I'm fine with adding Java 12 (current version), but I think we should also still have Java 11 ([an LTS, like Java 8](https://en.wikipedia.org/wiki/Java_version_history)).

Or do you feel Java 11 is either not that popular or sufficiently similar to Java 12 for it to be unnecessary?

PR #87, opened 2019-04-25
[python] remove user site from Dockerfile

AlexTugarev:
this will be automatically set in the Gitpod layer

also, this shouldn't prevent modules from being installed in custom images based on workspace-full.

Fixes https://github.com/gitpod-io/gitpod/issues/517

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -147,9 +147,9 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
     && pip3 install --upgrade pip \
     && pip3 install virtualenv pipenv pylint rope flake8 autopep8 pep8 pylama pydocstyle bandit python-language-server[all]==0.25.0 \
     && rm -rf /tmp/*
-# use global installation location for pip modules
-ENV PYTHONUSERBASE=/workspace/.pip-modules \
-    PIP_USER=yes
+# Gitpod will automatically add user site under `/workspace` to persist your packages.
+# ENV PYTHONUSERBASE=/workspace/.pip-modules \
+#    PIP_USER=yes
 
 ### Ruby ###
 RUN curl -sSL https://rvm.io/mpapis.asc | gpg --import - \


full/Dockerfile
```diff
@@ -147,9 +147,9 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
     && pip3 install --upgrade pip \
     && pip3 install virtualenv pipenv pylint rope flake8 autopep8 pep8 pylama pydocstyle bandit python-language-server[all]==0.25.0 \
     && rm -rf /tmp/*
-# use global installation location for pip modules
-ENV PYTHONUSERBASE=/workspace/.pip-modules \
-    PIP_USER=yes
+# Gitpod will automatically add user site under `/workspace` to persist your packages.
+# ENV PYTHONUSERBASE=/workspace/.pip-modules \
+#    PIP_USER=yes
```
jankeromnes: Nit: Why are you leaving the code commented-out here? As a sort of documentation/warning to users who'd be tempted to set `PYTHONUSERBASE` to something else?

AlexTugarev: yes, that was the idea. would you like to have this removed, but leaving the comment line?

AlexTugarev:
>  the modules installed by both the Docker images and the Gitpod user will co-exist peacefully?

exactly!

full/Dockerfile
jankeromnes: I don't really mind either way. I think this may be overly cautious, but it doesn't hurt to have an extra comment in the Dockerfile, so fine either way. 😄 👍

AlexTugarev:
thanks @jankeromnes! going to test with https://github.com/AlexTugarev/test.repo/blob/testing-python/.gitpod.dockerfile before merging.

`RUN unset PYTHONUSERBASE PIP_USER && pip3 install ...` is a workaround for custom images for the time being.

AlexTugarev:
tested with a custom workspace image inheriting `FROM gitpod/workspace-full:branch-at-remove-pip-user` and installing additional packages.

PR #92, opened 2019-05-10
[full] install netcat

jankeromnes:
This should allow doing things like `while ! nc -z localhost $1 ; do sleep 1 ; done` in `.gitpod.yml` tasks in order to wait for a port to open.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -13,6 +13,7 @@ RUN yes | unminimize \
         locales \
         man-db \
         nano \
+        netcat \
         software-properties-common \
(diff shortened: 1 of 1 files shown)


jankeromnes:
The preferred way to achieve my use case would be to parse `cat /proc/net/tcp /proc/net/tcp6`, so I'll close this pull request for now.

However, we're happy to install `netcat` if it can be useful to others. If you'd like to have `netcat` installed by default on Gitpod, please comment here, and I'll re-open this pull request.

PR #95, opened 2019-05-20
improvements to MySQL and Apache

meysholdt:
[Apache] enable module "rewrite"
[MySQL] introduce lock to prevent conflicting launches


meysholdt:
tested; works as intended.

PR #100, opened 2019-08-28
Add support for MySQL 8.0

eryjus:
Adds a MySQL 8.0 Gitpod.

I am looking for some help on how to test this properly since I believe there needs to be an image on DockerHub to back this commit at the same time -- I am not certain what the next steps are.

Any help in the right direction will be appreciated.

### DIFF
--- a/mysql-8.0/Dockerfile
+++ b/mysql-8.0/Dockerfile
@@ -1,0 +1,73 @@
+FROM    gitpod/workspace-full
+
+
+##
+## -- Set up the environment for a Docker build
+##    -----------------------------------------
+USER    root
+RUN     apt-get update 
+RUN     apt-get install debconf-utils -y -q
+ENV     DEBIAN_FRONTEND=noninteractive
+
+
+##
+## -- Install the Oracle Repo for MySQL
+##    ---------------------------------
+USER    root
+RUN     wget https://dev.mysql.com/get/mysql-apt-config_0.8.13-1_all.deb -O /usr/mysql-apt-config_0.8.13-1_all.deb \
+  &&    echo 4 | dpkg -i /usr/mysql-apt-config_0.8.13-1_all.deb \
+  &&    rm /usr/mysql-apt-config_0.8.13-1_all.deb
+RUN     apt-get update 
+
+
+##
+## -- Install MySQL server on top of the basic Docker Container
+##    ---------------------------------------------------------
+USER    root
+RUN     apt-get install mysql-server mysql-shell -y -q
+
+
+##
+## -- Freshen up all the other packages just in case
+##    ----------------------------------------------
+USER    root
+RUN     apt-get upgrade -y -q
+
+
+##
+## -- Now, we need to prepare for MySQL
+##    ---------------------------------
+USER    root
+RUN     mkdir -p /var/run/mysqld 
+RUN     mkdir -p /var/lib/mysql-upgrade
+RUN     mkdir -p /usr/share/mysql
+RUN     chown -R gitpod:gitpod /etc/mysql 
+RUN     chown -R gitpod:gitpod /var/run/mysqld 
+RUN     chown -R gitpod:gitpod /var/log/mysql 
+RUN     chown -R gitpod:gitpod /var/lib/mysql 
+RUN     chown -R gitpod:gitpod /var/lib/mysql-files 
+RUN     chown -R gitpod:gitpod /var/lib/mysql-keyring 
+RUN     chown -R gitpod:gitpod /var/lib/mysql-upgrade
+RUN     chown -R gitpod:gitpod /usr/share/mysql 
+COPY    mysql.cnf /etc/mysql/mysql.conf.d/mysqld.cnf
+COPY    client.cnf /etc/mysql/mysql.conf.d/client.cnf
+COPY    launch.sh /etc/mysql/mysql-bashrc-launch.sh
+RUN     chmod +x /etc/mysql/mysql-bashrc-launch.sh
+
+
+##
+## -- Set mysql to start under the gitpod user
+##    ----------------------------------------
+USER gitpod
+RUN echo "/etc/mysql/mysql-bashrc-launch.sh" >> ~/.bashrc
+
+
+##
+## -- clean up and give control back for the IDE to run
+##    -------------------------------------------------
+USER    root
+RUN     apt-get clean
+RUN     rm -rf /var/cache/apt/* /var/lib/apt/lists/* /tmp/* 
+USER    root
+
+ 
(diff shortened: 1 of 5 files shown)


jankeromnes:
Thanks for the Pull Request @eryjus! Our automated CircleCI Docker build didn't trigger here, because this PR is from a fork, but normally that's what you can use to test your Dockerfiles.

I'll trigger the build for you (but I may have to import your commit into a branch here to do that).

mysql-8.0/Dockerfile
```diff
@@ -0,0 +1,73 @@
+FROM    gitpod/workspace-full
+
+
+##
+## -- Set up the environment for a Docker build
+##    -----------------------------------------
+USER    root
+RUN     apt-get update 
+RUN     apt-get install debconf-utils -y -q
```
jankeromnes: Small note in passing, every `RUN` instruction creates a new Docker layer that can be cached.

With `apt-get update` and `apt-get intall`, it's important to have both in the same `RUN` command. Because otherwise what can happen is that the result of `apt-get update` gets cached in an old Docker layer, and if you modify the `apt-get install` instruction (e.g. by adding a package), it might try to use an out-dated list.

What we generally do is:

```Dockerfile
USER gitpod
RUN sudo apt-get update \
 && sudo apt-get install -y \
    package \
 && sudo rm -rf /var/lib/apt/lists/*
```

That way, in a single layer, you download an up-to-date apt list, then install what you need, and then remove the unnecessary apt list again before saving the resulting file system into a Docker layer (thus saving valuable disk space in your image).

mysql-8.0/Dockerfile
```diff
@@ -0,0 +1,73 @@
+FROM    gitpod/workspace-full
+
+
+##
+## -- Set up the environment for a Docker build
+##    -----------------------------------------
+USER    root
+RUN     apt-get update 
+RUN     apt-get install debconf-utils -y -q
+ENV     DEBIAN_FRONTEND=noninteractive
```
jankeromnes: Another note, `ENV` instruction will get persisted in the image, forever. This is not what you want, because once you have a container with a Terminal you can interact with, it will still believe that the frontend is non-interactive.

The solution is to define it just for the `apt-get` commands, like so:

```Dockerfile
USER gitpod
RUN sudo apt-get update \
 && DEBIAN_FRONTEND=noninteractive sudo apt-get install -y \
    package \
 && sudo rm -rf /var/lib/apt/lists/*
```

mysql-8.0/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+USER    root
+RUN     apt-get update 
+RUN     apt-get install debconf-utils -y -q
+ENV     DEBIAN_FRONTEND=noninteractive
+
+
+##
+## -- Install the Oracle Repo for MySQL
+##    ---------------------------------
+USER    root
```
jankeromnes: Note: No need to re-specify `USER` every time. Basically, in a Docker context, you have the same `USER` for all the instructions. You can change it mid-way e.g. with `USER gitpod`, and then it stays `gitpod` for all the rest of the instructions below.

jankeromnes:
Build triggered in #101. However, I notice that it was triggered here as well, but in theory it should fail after the first image because it won't be able to push to Docker Hub from a fork.

eryjus:
@jankeromnes, thank you for all the feedback.  Since this is my very first PR ever and the first time I am working with a Dockerfile, I am learning.  I appreciate your patience.

So, I want to do the right thing here.  It appears that you moved my changes to a branch in this repo (rather than a fork).  I believe that this PR should be closed (well, at the right time) and all the suggestions from this PR really apply to #101 and therefore need to be updated there.

Before this gets closed, I also want to make sure we conclude all the different conversation points.

jankeromnes:
> Since this is my very first PR ever and the first time I am working with a Dockerfile, I am learning. I appreciate your patience.

Wow, congratulations on your first PR, and thanks a lot for your contribution! 🙂 It is much appreciated. 💯

I didn't realize. If you have any question regarding all this (Gitpod, Pull Requests, Dockerfiles...) or if anything seems confusing / doesn't make sense to you, please always feel free to ask. 🙂

> Before this gets closed, I also want to make sure we conclude all the different conversation points.

This sounds like a good plan. This Pull Request can remain open in the meantime. 👍

eryjus:
> If you have any question regarding all this (Gitpod, Pull Requests, Dockerfiles...)...

OK, so I thought that `sudo` was not available in Gitpod (and I can understand why)....  Perhaps that feature is removed when adding the Gitpod layer and it is really available during the Docker build?

eryjus:
So, I fetched out the `mysql-8.0` branch from this repo and made my changes there.  I wanted to push that back, but I do not have write access.  I have probably made a complete mess of this, and I'm not sure what the next step on my part would be.

jankeromnes:
> OK, so I thought that sudo was not available in Gitpod (and I can understand why).... Perhaps that feature is removed when adding the Gitpod layer and it is really available during the Docker build?

Yes, this is exactly correct. 🙂 Also, we'd love to enable it in running workspaces one day, but there are still a few security concerns that need to be cleared first.

meysholdt:
Can this replace the old MySQL 7.0 image? I don't think we need to have two MySQL images.

eryjus:
@meysholdt, @jankeromnes raised a very similar question [here](https://github.com/gitpod-io/workspace-images/pull/101#issuecomment-526060442).   I certainly do not have a problem either way, but in my opinion there are a few things to consider when evaluating this decision.

- The official image is MySQL version 5.7.  This is the most recent version that is available in the Ubuntu Repo.
- Oracle has jumped from version 5.7 to version 8.0.  Included in this version by default is new XDevAPI which is a new API interface -- it is this feature, by the way, I am striving to be able to play with.
- MySQL 8.0 is installed from the Oracle Repo (some may care about this given how Oracle has re-licensed and started charging for updates to what might be the most popular programming language in the world).
- MariaDB (forked from MySQL from version 5.5 I believe) is not expected to maintain compatibility with MySQL going forward.

jankeromnes:
Hi @eryjus! Sorry again for lagging on this, but in a fortunate turn of event, our recent upgrade to Ubuntu 20.04 also upgraded MySQL to version 8! 🎉 

```
gitpod /workspace/gitpod-mysql-workspace $ mysql --version
mysql  Ver 8.0.19-0ubuntu4 for Linux on x86_64 ((Ubuntu))

gitpod /workspace/gitpod-mysql-workspace $ mysqld --version
/usr/sbin/mysqld  Ver 8.0.19-0ubuntu4 for Linux on x86_64 ((Ubuntu))
```

(However, it seems that our `mysql.cnf` is now incompatible with MySQL 8, as reported in https://github.com/gitpod-io/gitpod/issues/1399 -- so we're now trying to update it as well.)

Meanwhile, I believe that the original feature request has been implemented. 😇 🎉

PR #101, opened 2019-08-29
Add support for MySQL 8.0

jankeromnes:
Imported the commit from #100 into a repo branch in order to trigger a CircleCI build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,6 +103,20 @@ jobs:
             ./.circleci/build_image.sh mysql/Dockerfile gitpod/workspace-mysql
           no_output_timeout: 30m
 
+  workspace-mysql-8:
+    docker:
+      - image: docker:stable
+    steps:
+      - checkout
+      - setup_remote_docker
+      - attach_workspace:
+          at: .
+      - run:
+          command: |
+            docker load -i workspace-full.tar
+            ./.circleci/build_image.sh mysql-8.0/Dockerfile gitpod/workspace-mysql-8
+          no_output_timeout: 30m
+
   workspace-mongodb:
     docker:
       - image: docker:stable
@@ -151,6 +165,9 @@ workflows:
           requires:
             - workspace-full
       - workspace-mysql:
+          requires:
+            - workspace-full
+      - workspace-mysql-8:
           requires:
             - workspace-full
       - workspace-mongodb:
--- a/mysql-8.0/Dockerfile
+++ b/mysql-8.0/Dockerfile
@@ -1,0 +1,73 @@
+FROM    gitpod/workspace-full
+
+
+##
+## -- Set up the environment for a Docker build
+##    -----------------------------------------
+USER    root
+RUN     apt-get update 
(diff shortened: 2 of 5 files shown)


jankeromnes:
I've also added a commit to hook up the new Dockerfile properly so that CircleCI actually builds it.

jankeromnes:
Also, I'm not yet sure whether we want to officially support MySQL 8.0 in a core Docker Hub image.

An alternative way would be for @eryjus to add this Dockerfile to one of their projects, and hook it up with a `.gitpod.yml` like so:

```yaml
image:
  file: Dockerfile
```

jankeromnes:
The build failure is unrelated to the MySQL 8.0 change:

```
E: Unable to locate package clang-format-9
E: Unable to locate package clang-tidy-9
E: Unable to locate package clang-tools-9
E: Unable to locate package lld-9
```

It seems like @Sylvestre changed something in apt.llvm.org yet again. 😛

sylvestre:
yeah, we are working on a new version of llvm, so we branched
As you are probably using the master repo, we moved to 10.

Two potential fixes:
* you use llvm-toolchain-9 (the qualification branch)
* you install clang, lld, etc
it should default to the correct version

jankeromnes:
Wow, there is a Clang 10 now! ✨

Switched to the qualification branch in #102.

Also, I agree that we should probably switch to `clang`, `lld`, etc. which seems more "standard". Will work on that soon.

Thanks @sylvestre! 👍

sylvestre:
De rien @jankeromnes 
I just documented it here: https://apt.llvm.org/index.php#default_pkg

jankeromnes:
@sylvestre but will you always get same version with `apt-get install clang`, regardless of which repository you use? Or does it change, e.g.:
- `deb http://apt.llvm.org/disco/ llvm-toolchain-disco main` + `apt-get install clang` → Clang 10?
- `deb http://apt.llvm.org/disco/ llvm-toolchain-disco-8 main` + `apt-get install clang` → Clang 8?`
- `deb http://apt.llvm.org/disco/ llvm-toolchain-disco-9 main` + `apt-get install clang` → Clang 9?

jankeromnes:
(Rebased on top of the `llvm-toolchain-disco-9` fix.)

sylvestre:
> @sylvestre but will you always get same version with `apt-get install clang`, regardless of which repository you use? Or does it change, e.g.:
> 
>     * `deb http://apt.llvm.org/disco/ llvm-toolchain-disco-8 main` + `apt-get install clang` → Clang 8?`

I am not sure it is really interesting: if you select a specific branch, it is because you have a good reason, so, you can hardcode the version.
Currently, clang isn't available for -8 or -9

meysholdt:
Can this replace the old MySQL 7.0 image? I don't think we need to have two MySQL images.

jankeromnes:
Gitpod has now been upgraded to MySQL 8! 🎉 Resolved fixed. https://github.com/gitpod-io/workspace-images/pull/100#issuecomment-614633230

PR #105, opened 2019-09-03
[python] bump versions of pyenv installations

AlexTugarev:
the selection game implemented in the ms-python extension is won by the highest version number, so let's bump it to the latest.


AlexTugarev:
I tested it with https://github.com/AlexTugarev/django-locallibrary-tutorial/blob/test8/.gitpod.yml#L2 and it worked nicely.

PR #110, opened 2019-09-09
[full] Fix #109 by using 'apt -t unstable'

### LINKED ISSUE #109: clangd-9 : Breaks: clang-tools-9 (< 1:9~+rc1-1~exp2) but 1:9~svn371301-1~exp1~20190908074810.48 is to be installed
The command:

https://github.com/gitpod-io/workspace-images/blob/ce2f3e0d5f99b112eb3dbb9210db641f6f793fcc/full/Dockerfile#L34-L43

currently fails with:

```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

 clangd-9 : Breaks: clang-tools-9 (< 1:9~+rc1-1~exp2) but 1:9~svn371301-1~exp1~20190908074810.48 is to be installed
```

I don't understand what this means, but I'll try to remove `clang-tools-9`, because `clangd-9` is more important to us.

Cc @sylvestre, as maybe this is a apt.llvm.org bug.

jankeromnes:
I have no idea what I'm doing.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -36,7 +36,6 @@ RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - \
     && apt-get install -yq \
         clang-format-9 \
         clang-tidy-9 \
-        clang-tools-9 \
         clangd-9 \
         cmake \
         gdb \


jankeromnes:
Failed with:

```
E: The value 'unstable' is invalid for APT::Default-Release as such a release is not available in the sources
```

jankeromnes:
No trying without `clang-tools-9`, which seems problematic and maybe not actually useful.

jankeromnes:
Oh, the problem actually persists even without `clang-tools-9`:

```bash
apt-get install -yq \
        clang-format-9 \
        clang-tidy-9 \
        clang-tools-9 \
        clangd-9 \
        cmake \
        gdb \
        lld-9
```

```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

 clangd-9 : Breaks: clang-tools-9 (< 1:9~+rc1-1~exp2) but 1:9~svn371301-1~exp1~20190908074810.48 is to be installed
```

jankeromnes:
Doesn't work.

PR #111, opened 2019-09-09
Upgrade all images to Clang 10

### LINKED ISSUE #109: clangd-9 : Breaks: clang-tools-9 (< 1:9~+rc1-1~exp2) but 1:9~svn371301-1~exp1~20190908074810.48 is to be installed

jankeromnes:
This should work around or fix #109.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -32,16 +32,16 @@ RUN { echo && echo "PS1='\[\e]0;\u \w\a\]\[\033[01;32m\]\u\[\033[00m\] \[\033[01
 
 ### C/C++ ###
 RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - \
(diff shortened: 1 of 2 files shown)


jankeromnes:
All green!

Screenshot 2019-09-09 at 14 19 17

PR #114, opened 2019-09-26
[WIP] Try to use the generic name

sylvestre:


jankeromnes:
Super cool! Many thanks for your pull request @sylvestre! 💯

For context, this change depends on https://bugs.llvm.org/show_bug.cgi?id=43466, which should be fixed now.

Also, CircleCI is asleep, I'll try to wake it up.

jankeromnes:
Woohoo, all green! 🎉

Screenshot 2019-09-27 at 08 23 48

Not using version suffixes is a much better approach going forward, as we won't ever have to update them everywhere with every LLVM release.

Thanks again! 💯

PR #118, opened 2019-09-30
use avdmanager to preload Android VM

sr229:
Unfortunately because of the first Flutter PR's shortcomings I still have to create this PR, which is a follow up.

This PR leverages `avdmanager`, which is a CLI tool to make AVD creation more tolerable on non-interactive environments.

the Gitpod AVD I preloaded comes with Android API Level 28 on x86. Name of the AVD is `gitpod`.


sr229:
We're not testing flutter apparently, is this intended?

sr229:
Going to close this down so I can merge this to the android dockerfile instead.

PR #119, opened 2019-10-01
[full] Install tk-dev to help Python work with graphical environments

jankeromnes:
`tk-dev` needs to be installed _before_ `pyenv` installs Python, thus I install it in `gitpod/workspace-full` and not just `gitpod/workspace-full-vnc`.

See also https://github.com/pyenv/pyenv/issues/94 and https://github.com/pyenv/pyenv/issues/1375.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -172,7 +172,9 @@ ENV PATH=/home/gitpod/.nvm/versions/node/v${NODE_VERSION}/bin:$PATH
 
 ### Python ###
 ENV PATH=$HOME/.pyenv/bin:$HOME/.pyenv/shims:$PATH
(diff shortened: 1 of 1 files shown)


jankeromnes:
Success!

Screenshot 2019-10-01 at 16 37 02

meysholdt:
does this pull X11 into workspace-full?

jankeromnes:
Let's close this for now.

If you think it would make sense to have `tk-dev` in Gitpod's VNC image, please re-open this or file an issue. Thanks!

felipecassiors:
It makes a lot of sense to have the `tk-dev` in the VNC image because for people who need, the solution is tricky.

PR #117, opened 2019-09-30
[full] Upgrade Node.js from 10.15.3 → 10.16.3 LTS

jankeromnes:

PR #121, opened 2019-10-24
Build and test workspace-full using dazzle

32leaves:
`workspace-full` is a big image (> 8GiB) but with largely independent layers. Changing one layer (e.g. #117) needlessly invalidates others. 

[Dazzle](https://github.com/32leaves/dazzle) is an experimental Docker image builder that can build independent layers/"chunks". To ensure that those layers do not interfere with each other and the their merging has no adverse effects, dazzle supports testing the image.

This PR builds `workspace-full` using dazzle which should reduce workspace startup time variance considerably as the chances of a cache-miss are reduced. The image built using dazzle passes the same tests as the original image does (tests are part of this PR).

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -17,7 +17,11 @@ RUN yes | unminimize \
         vim \
         multitail \
     && locale-gen en_US.UTF-8 \
+    && mkdir /var/lib/apt/dazzle-marks \
+    && apt-mark showmanual >> /var/lib/apt/dazzle-marks/base.manual.list \
+    && apt-mark showauto >> /var/lib/apt/dazzle-marks/base.auto.list \
     && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/*
+
 ENV LANG=en_US.UTF-8
 
 ### Gitpod user ###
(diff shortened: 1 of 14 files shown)


full/Dockerfile
```diff
@@ -30,21 +30,34 @@ WORKDIR $HOME
...
+LABEL dazzle/layer=lang-c
+LABEL dazzle/test=tests/lang-c.yaml
+USER root
 RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - \
     && apt-add-repository -yu "deb http://apt.llvm.org/disco/ llvm-toolchain-disco main" \
     && apt-get install -yq \
         clang-format \
         clang-tidy \
-        clang-tools \
+        # clang-tools \ # breaks the build atm
```
geropl: Is this a test leftover?

32leaves: No. There's an issue with the clang apt sources/packages at the moment. This is independent of this PR/dazzle. The package itself just doesn't install at the moment.

```
Unpacking clangd (1:10.0-48+0~20191017130034.13~1.gbp24c8a8) ...
dpkg: error processing archive /tmp/apt-dpkg-install-kqkfT1/26-clangd_1%3a10.0-48+0~20191017130034.13~1.gbp24c8a8_amd64.deb (--unpack):
 trying to overwrite '/usr/bin/clangd', which is also in package clang-tools 1:8.0-48~exp1ubuntu1
```

32leaves:
Fixes #122
Correctly merges the dpkg status files which avoids duplicate installation of the same packages.

PR #129, opened 2019-11-07
add nim

JesterOrNot:


JesterOrNot:
@RDIL any thoughts?

RDIL:
I don't see any issues

svenefftinge:
The size of our workspace-full is already quite big. The intend of it is to provide a good default container that works for 80% of all projects. Although we have included the most popular programming languages we should avoid adding more niche languages because it will increase size and complexity further.

JesterOrNot:
Ok!

PR #130, opened 2019-11-07
Place stuff in .bashrc, not .bash_profile

32leaves:
fixes Java, Ruby and Rust in non-login shells (like the ones opened in GItpod).

Try here: https://gitpod.io/#https://github.com/32leaves/test-repo/tree/test-wsfull-sdkfix

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -37,9 +37,9 @@ RUN { echo && echo "PS1='\[\e]0;\u \w\a\]\[\033[01;32m\]\u\[\033[00m\] \[\033[01
 USER gitpod
 # use sudo so that user does not get sudo usage info on (the first) login
 RUN sudo echo "Running 'sudo' for Gitpod: success"
-# create .profile.d folder and source it in the bashrc
-RUN mkdir /home/gitpod/.profile.d && \
-    echo "for i in \$(ls \$HOME/.profile.d/*); do source \$i; done" >> /home/gitpod/.profile
+# create .bashrc.d folder and source it in the bashrc
+RUN mkdir /home/gitpod/.bashrc.d && \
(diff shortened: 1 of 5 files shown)


full/Dockerfile
```diff
@@ -253,11 +255,15 @@ RUN sudo apt-get update \
...
-RUN curl -fsSL https://sh.rustup.rs | sh -s -- -y \
+RUN cp /home/gitpod/.profile /home/gitpod/.profile_orig && \
+    curl -fsSL https://sh.rustup.rs | sh -s -- -y \
     && .cargo/bin/rustup update \
     && .cargo/bin/rustup component add rls-preview rust-analysis rust-src \
     && .cargo/bin/rustup completions bash | sudo tee /etc/bash_completion.d/rustup.bash-completion > /dev/null \
-    && .cargo/bin/rustup target add x86_64-unknown-linux-musl
+    && .cargo/bin/rustup target add x86_64-unknown-linux-musl \
+    && grep -v -F -x -f /home/gitpod/.profile_orig /home/gitpod/.profile > /home/gitpod/.bashrc.d/80-rust
+    # && mv /home/gitpod/.profile_orig /home/gitpod/.profile
```
geropl: Leftover?

geropl:
@32leaves Could you elaborate why the move from `profile` to `bashrc` is necessary? Just trying to understand.

AlexTugarev:
> Could you elaborate why the move from profile to bashrc is necessary? Just trying to understand.

@geropl, in Theia we're starting `bash -i` (not `bash -l`) for new Terminals which by default will only source `.bashrc`, but not `.profile`. (`.profile` again sources `.bashrc`)

full/Dockerfile
32leaves: :+1:

geropl:
@AlexTugarev Thank you for the explanation.

PR #131, opened 2019-11-08
Bump rust version to v1.39.0

geropl:


JesterOrNot:
looks good

full/Dockerfile
```diff
@@ -257,7 +257,7 @@ RUN sudo apt-get update \
 
 RUN cp /home/gitpod/.profile /home/gitpod/.profile_orig && \
     curl -fsSL https://sh.rustup.rs | sh -s -- -y \
-    && .cargo/bin/rustup update \
+    && .cargo/bin/rustup install 1.39.0 \
```
jankeromnes: Wouldn't `rustup update` already pick `1.39.0` today?

Not a big issue, but I'm a bit worried that by pinning the version here, we're committing to keeping Rust up-to-date manually forever.

jankeromnes:
Testing the CircleCI PR image with https://github.com/jankeromnes/nushell/commit/bcdedc90f7fb62ebc22bb83bcad6da2197f1e4a0

full/Dockerfile
geropl: @jankeromnes  Yes, it would - if it would not be for Dazzle. If the layer description does not change the layer is taken from the last build.

We could also introduce pseudo-counters or sth. to trigger a re-build of the layer. WDYT?

jankeromnes: > We could also introduce pseudo-counters or sth. to trigger a re-build of the layer. WDYT?

This sounds like a forever-manual-update that is just a bit easier to trigger ("increment counter" instead of "hunt for the correct version string"). So, a little better, but still not ideal?

geropl: Ok. Will leave it as is for now.

PR #141, opened 2019-12-16
Ensure rubygems doesn't break

RDIL:
Due to an update with Bundler, this is needed for ruby projects not to break

cc @jankeromnes for testing

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -225,6 +225,9 @@ RUN curl -sSL https://rvm.io/mpapis.asc | gpg --import - \
         && rvm install 2.6 \
         && rvm use 2.6 --default \
         && rvm rubygems current \
+        && gem update --system \
+        && gem update \
+        && gem uninstall rubygems-update \
         && gem install bundler --no-document \
         && gem install solargraph --no-document" \
     && echo '[[ -s "$HOME/.rvm/scripts/rvm" ]] && source "$HOME/.rvm/scripts/rvm" # Load RVM into a shell session *as a function*' >> /home/gitpod/.bashrc.d/70-ruby


jankeromnes:
Interesting, thanks a lot for this quick contribution @RDIL!

Do you have more information on how projects break (e.g. what the symptoms or error messages are) and where this solution comes from (e.g. official bug report or stackoverflow reply)?

RDIL:
@jankeromnes the bundler team responded to my support request: https://github.com/bundler/bundler/issues/7488 - we use [RVM](https://rvm.io) which adds in a Gem to the system store ([RubyGems-Bundler](https://github.com/rvm/rubygems-bundler)) - the only problem is that RubyGems updated, which now automatically installs this bundler update, which in turn gets installed and RVM attempts to patch it using the above stated Gem, the only problem is that *that gem doesn't work anymore*, meaning bundler is broken because it is stuck mid patch.

RDIL:
@jankeromnes this breaks dependency management, I can help out with any testing, just need to get this in so I can teach tonight

RDIL:
Also why is everything suddenly light theme???

jankeromnes:
> @jankeromnes this breaks dependency management, I can help out with any testing, just need to get this in so I can teach tonight

@RDIL Oh, I'm so sorry that I missed this. Seemed like the update was time-critical. 😕 

I hope you were able to get around this, e.g. by setting up a custom `.gitpod.Dockerfile` for your project with `gp init`.

jankeromnes:
> Also why is everything suddenly light theme???

We wanted to improve the look-and-feel of Gitpod, and also make it configurable. Currently, you can switch the IDE and the dashboard app to dark theme, but the www.gitpod.io website doesn't have a dark theme yet.

RDIL:
Yes, it can be worked around, but this is still a problem on workspace full

PR #143, opened 2019-12-17
Full: update gradle to 6.0.1

RDIL:
Admittedly this is just to get the SHAs to change so a newer version of gradle can be installed, but it works.


jankeromnes:
Thanks a lot! Looks good to me. I also snuck in a minor Java version change in order to restore JavaFX support.

jankeromnes:
Wasn't able to test my sneaked-in change, but your change successfully upgraded Gradle. Merging this. 🙂

PR #144, opened 2019-12-27
Add Gitpod config

jcmarcfloress:
Thanks


.gitpod.yml
```diff
@@ -0,0 +1 @@
+{}
```
jankeromnes: This doesn't look like a valid Gitpod configuration.

README.md
```diff
@@ -1,3 +1,5 @@
+[![Gitpod Ready-to-Code](https://img.shields.io/badge/Gitpod-Ready--to--Code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/gitpod-io/workspace-images)
```
jankeromnes: I don't think we want to have this badge all the way at the top of the README.

PR #145, opened 2020-01-04
Bump python to 3.7.6

RDIL:
Bug fix release, no functional changes


RDIL:
I would do that, but tensorflow still hasn't added the code for Py3.8 meaning it will break your test repo

jankeromnes:
Thanks a lot for this valuable insight @RDIL. I think tensorflow support is a good enough reason to hold off on the 3.8 upgrade for now.

PR #146, opened 2020-01-07
[full] Install tk-dev to support some graphical environments in Python

jankeromnes:
This is a revival of #119:

> `tk-dev` needs to be installed _before_ `pyenv` installs Python, thus I install it in `gitpod/workspace-full` and not just `gitpod/workspace-full-vnc`.
> 
> See also pyenv/pyenv#94 and pyenv/pyenv#1375.

Needed by:
- https://github.com/JesterOrNot/pycalc
- https://community.gitpod.io/t/how-to-install-tk-and-tkinter-gitpod/274

However, as @meysholdt correctly pointed out, this seems to add `libx11` in all workspaces (even non-graphical).

We should check what the impact is of installing `tk-dev` on the `gitpod/workspace-full` image size.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -195,7 +195,9 @@ LABEL dazzle/layer=lang-python
 LABEL dazzle/test=tests/lang-python.yaml
 USER gitpod
 ENV PATH=$HOME/.pyenv/bin:$HOME/.pyenv/shims:$PATH
-RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash \
+RUN sudo apt-get update -q \
+    && sudo apt-get install -yq tk-dev \
+    && curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash \
     && { echo; \
         echo 'eval "$(pyenv init -)"'; \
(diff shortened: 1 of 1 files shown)


jankeromnes:
Failed with:

```
              Setting up tzdata (2019c-0ubuntu0.19.04) ...
              debconf: unable to initialize frontend: Dialog
              debconf: (Dialog frontend will not work on a dumb terminal, an emacs shell buffer, or without a controlling terminal.)
              debconf: falling back to frontend: Readline
              Configuring tzdata
              ------------------
              
              Please select the geographic area in which you live. Subsequent configuration
              questions will narrow this down by presenting a list of cities, representing
              the time zones in which they are located.
              
                1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
                2. America     5. Arctic     8. Europe    11. SystemV
                3. Antarctica  6. Asia       9. Indian    12. US

Too long with no output (exceeded 30m0s): context deadline exceeded
```

This needs to be done non-interactively because of Docker.

RDIL:
This should be made part of #194 because of all the python changes

jankeromnes:
Superseded by https://github.com/gitpod-io/workspace-images/pull/303 whicht attempts to make `tk-dev` opt-in.

PR #152, opened 2020-01-14
Changes in full image

RDIL:
* Python: Move several packages from 2 to 3 (prevents conflicts)
* Shrink image (replace asciidoctor with some of its dependencies, zip and unzip, minimizing the package count)


jankeromnes:
Hey, thanks a lot for these upgrades! Will try to review soon, unless someone else gets to it first.

Random thought, I think at some point we should make Python 3 the default version. It was already like that in the past, and now that 2 is deprecated, I think we can still keep it, but it shouldn't probably remain the default very long.

RDIL:
Not too random

full/Dockerfile
```diff
@@ -228,6 +233,7 @@ RUN curl -sSL https://rvm.io/mpapis.asc | gpg --import - \
         rvm requirements \
         && rvm install 2.5 \
```
JesterOrNot: If we are adding 2.7 let's remove 2.5 for the sake of image size

RDIL: https://www.jetbrains.com/lp/devecosystem-2019/ruby/

JesterOrNot: Fair enough

JesterOrNot: @jankeromnes is it ok adding an additional Ruby version?

RDIL: You know what, I'm just going to remove some useless stuff from the image hang on

RDIL: Ok so the image should be much smaller now (the logs too, I fixed an issue where it was more than 40000 lines long for no real purpose)

RDIL:
@jankeromnes @JesterOrNot can we please get this merged?

PR #156, opened 2020-02-06
Update Rust installation in workspace-full

coolreader18:
`rustup install` as an alias for `rustup update` was [deprecated recently](https://users.rust-lang.org/t/notice-deprecation-of-rustup-install-and-rustup-uninstall/35491), and the `rls-preview` feature was [renamed to `rls`](https://github.com/rust-lang/rls/pull/1208) with Rust 2018.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -245,7 +245,7 @@ RUN sudo apt-get update \
 
 RUN cp /home/gitpod/.profile /home/gitpod/.profile_orig && \
     curl -fsSL https://sh.rustup.rs | sh -s -- -y \
-    && .cargo/bin/rustup update stable \
+    && .cargo/bin/rustup toolchain install 1.41 \
     && .cargo/bin/rustup component add rls rust-analysis rust-src \
     && .cargo/bin/rustup completions bash | sudo tee /etc/bash_completion.d/rustup.bash-completion > /dev/null \
     && .cargo/bin/rustup target add x86_64-unknown-linux-musl \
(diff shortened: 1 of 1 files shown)


geropl:
Hi @coolreader18,

thx for your contribution.

```
rustup update stable
```
We like to use specific version numbers because we're caching individual image layers (we use [dazzle](https://github.com/csweichel/dazzle)).

According to your link `rustup toolchain install 1.41.0` should do the trick.

coolreader18:
Would it be best to just make a PR every ~6 weeks when a new Rust version is released?

geropl:
> Would it be best to just make a PR every ~6 weeks when a new Rust version is released in order to make sure the toolchain is up to date?

Not exactly. Let me explain: The image that is build from this file is the default one-size-fits-all image that is used as basis for every gitpod.io workspace without special config. Here it is important to re-use the same layer as often as possible (to speed up workspaces starts) and not so much that each individual tool is updated to the latest version (because most users won't use/see it anyway).

If you as a user of gitpod.io want a specific/new version of a tool for a repository of yours, the best way to achieve that is by installing in your [repo-local gitpod config](https://www.gitpod.io/docs/config-docker/#creating-docker-images-for-gitpod).
Was that helpful?

By the way, I just created a [similar PR](https://github.com/gitpod-io/workspace-images/pull/157) with the goal to reduce the image size. I copied some of your suggestions as well.

jankeromnes:
Thanks a lot for your contribution @coolreader18! 🙏  🎉

It was merged as part of https://github.com/gitpod-io/workspace-images/pull/157 & 33188f9dc5b1d514c9901a03bc8ffdbf5f787a8a.

PR #157, opened 2020-02-06
Rust toolchain: Update to 1.41.0, uninstall stable, more tools

geropl:
This optimizes the Rust installation in several ways:
 - uninstalls one (unused) toolchain (saves **~1.7GB**)
 - updates the installed toolchain to 1.41.0
 - installs additional `cargo` plugins for convenience

Thanks to @coolreader18 for the hint with `rustup toolchain install` and deprecated `rls-preview`.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -245,13 +245,19 @@ RUN sudo apt-get update \
 
 RUN cp /home/gitpod/.profile /home/gitpod/.profile_orig && \
     curl -fsSL https://sh.rustup.rs | sh -s -- -y \
-    && .cargo/bin/rustup install 1.39.0 \
-    && .cargo/bin/rustup component add rls-preview rust-analysis rust-src \
+    && .cargo/bin/rustup toolchain install 1.41.0 \
+    && .cargo/bin/rustup default 1.41.0 \
+    # Save image size by removing now-redudant stable toolchain
+    && .cargo/bin/rustup toolchain uninstall stable \
+    && .cargo/bin/rustup component add \
+        rls \
+        rust-analysis \
+        rust-src \
     && .cargo/bin/rustup completions bash | sudo tee /etc/bash_completion.d/rustup.bash-completion > /dev/null \
     && .cargo/bin/rustup target add x86_64-unknown-linux-musl \
     && grep -v -F -x -f /home/gitpod/.profile_orig /home/gitpod/.profile > /home/gitpod/.bashrc.d/80-rust
 
-RUN bash -lc "cargo install cargo-watch"
+RUN bash -lc "cargo install cargo-watch cargo-edit cargo-tree"
 
 ### Prologue (built across all layers) ###
 LABEL dazzle/layer=dazzle-prologue


jankeromnes:
Awesome work, many thanks!

Looking forward to test this on a few Rust repos.

Also, I guess Gitpod users won't miss the pre-installed nightly, as it can be installed & switched to conveniently via Dockerfile anyway.

geropl:
> Also, I guess Gitpod users won't miss the pre-installed nightly, as it can be installed & switched to conveniently via Dockerfile anyway.

The current `gitpod/workspace-full` contains two toolchains, `stable` and `1.39.0` (where stable is to the latest stable rust version from the time the image was build). Those are nearly identical as we install versions only if they are stable.

jankeromnes:
Oh. That does indeed sound unnecessary. 😲

jankeromnes:
Interesting, the base image build worked, but all of the dependent images failed. (There is a warning about Git, and then an error `object not found`.)

Screenshot 2020-02-06 at 17 54 09

Might be an internal CircleCI bug though, and doesn't prevent us from testing the new base image. 🙂

jankeromnes:
Full CircleCI error for dependent image builds:

```
Either git or ssh (required by git to clone through SSH) is not installed in the image. Falling back to CircleCI's native git client but the behavior may be different from official git. If this is an issue, please use an image that has official git and ssh installed.
Enumerating objects: 9, done.
Counting objects: 100% (9/9), done.
Compressing objects: 100% (9/9), done.
Total 697 (delta 3), reused 4 (delta 0), pack-reused 688

object not found
```

EDIT: Ah, that wasn't the latest build. Your amended force-push seems to have re-triggered builds that work better:

Screenshot 2020-02-06 at 17 56 48

jankeromnes:
FYI, to test this in a Rust repo, you can use the following base image in a `.gitpod.yml`:

```yml
image: gitpod/workspace-full:branch-gpl-update-rust
```

JesterOrNot:
Hey, I don't think we should scrap stable, it might break docker builds

jankeromnes:
> Hey, I don't think we should scrap stable, it might break docker builds

Oops. Which Docker builds do you think will break?

JesterOrNot:
Wouldn't it break anyone's build that required stable?

JesterOrNot:
Wouldn't it break anyone's build that required stable?

jankeromnes:
> Wouldn't it break anyone's build that required stable?

It's my understanding that `1.41.0` is currently the stable version: https://www.rust-lang.org/

We're planning to update this version every time a new Rust version becomes stable.

Are you worried about general compatibility between projects and the Rust compiler we ship, or do you think that some projects might require an explicit `stable` version to exist, which would be missing from our image?

JesterOrNot:
Nevermind this seems good, my bad 👍

jankeromnes:
No worries, thanks a lot for chiming in! The more eyes we have on these updates, and the more worries we raise, the less risk we'll take of breaking user projects downstream.

JesterOrNot:
Is this ready to merge?

jankeromnes:
Yes, I think this is ready to merge.

(An earlier commit was actually merged 3 hours ago, then reverted. Assuming this was accidental, and ignoring since this predates the successful builds and approvals here.)

PR #160, opened 2020-02-07
Add PHP Extension: Bcmath and Ctype

vikas5914:
Both Bcmath and ctype are required for standard Laravel project.


vikas5914:
@jankeromnes I have created a Minimalist Laravel demo for gitpod.io. Its works out of the box without any docker file: https://github.com/vikas5914/gitpod-laravel-demo

PR #162, opened 2020-02-10
[WIP] Pin dependencies

JesterOrNot:
Pin dependencies to make images ephemeral see https://nickjanetakis.com/blog/docker-tip-18-please-pin-your-docker-image-versions

PR #163, opened 2020-02-12
Use latest Git instead of the old one provided by Ubuntu

felipecassiors:
This way the users will have the latest version of Git (currently 2.25) instead of the older one provided in the Ubuntu repositories (currently 2.20).

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -23,6 +23,11 @@ RUN yes | unminimize \
 
 ENV LANG=en_US.UTF-8
 
+### Git ###
+RUN add-apt-repository -y ppa:git-core/ppa \
+    && apt-get install -yq git \
+    && rm -rf /var/lib/apt/lists/*
+
 ### Gitpod user ###
 # '-l': see https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user
 RUN useradd -l -u 33333 -G sudo -md /home/gitpod -s /bin/bash -p gitpod gitpod \


full/Dockerfile
```diff
@@ -23,6 +23,11 @@ RUN yes | unminimize \
 
 ENV LANG=en_US.UTF-8
 
+### Git ###
+RUN add-apt-repository -y ppa:git-core/ppa \
+    && apt-get install -yq git \
+    && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/*
```
jankeromnes: Note: Only `rm -rf /var/lib/apt/lists/*` is required to clean up after `apt-get`.

Sources:
- https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get
- https://hackernoon.com/tips-to-reduce-docker-image-sizes-876095da3b34

jankeromnes:
Also, let's see what the build says. I think `git` is already part of our base image `buildpack-deps:disco`, so I wonder how `apt-get` will like the renewed `install` instruction.

full/Dockerfile
felipecassiors: Good to know, done!

jankeromnes:
FYI, the previous CircleCI build successfully installed the latest Git, so your Pull Request works as expected. 🎉 

```
The following additional packages will be installed:
                git-man
              Suggested packages:
                gettext-base git-daemon-run | git-daemon-sysvinit git-doc git-el git-email
                git-gui gitk gitweb git-cvs git-mediawiki git-svn
              The following packages will be upgraded:
                git git-man
              2 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
[...]
              Setting up git-man (1:2.25.0-1~ppa0~ubuntu19.04.1) ...
              Setting up git (1:2.25.0-1~ppa0~ubuntu19.04.1) ...
              Processing triggers for man-db (2.8.5-2) ...
```

However, the build later failed with:

```
              github.com/sourcegraph/go-langserver (download)
              get "golang.org/x/tools/cmd/gopls": found meta tag get.metaImport{Prefix:"golang.org/x/tools", VCS:"git", RepoRoot:"https://go.googlesource.com/tools"} at //golang.org/x/tools/cmd/gopls?go-get=1
              get "golang.org/x/tools/cmd/gopls": verifying non-authoritative meta tag
              package golang.org/x/tools/cmd/gopls: cannot find package "golang.org/x/tools/cmd/gopls" in any of:
              	/home/gitpod/go/src/golang.org/x/tools/cmd/gopls (from $GOROOT)
              	/home/gitpod/go-packages/src/golang.org/x/tools/cmd/gopls (from $GOPATH)
```

but that seems unrelated to your Pull Request (same thing happened in #162). Looks like our Go install instructions, which used to work before, are currently broken. We'll look into fixing this.

felipecassiors:
> Many thanks for this contribution! I agree that having the latest stable Git is a must for development.
> 
> (As a side-note, you could have also used `sudo apt-get remove git && brew install git` instead of the PPA, but I agree that using the PPA is preferable here because we want to keep the image size as small as possible and also we don't mind using more complicated/convoluted install commands.)

That's right, I forgot that `brew` is available in the image (but personally I have never used it). I'm used to install the latest Git in Ubuntu this way, both in [my Vagrant box](https://github.com/felipecassiors/ubuntu1804-4dev/blob/master/provision/additional-tools.sh#L16) and in [my Jenkins Agent docker image](https://github.com/felipecassiors/jenkins-agent-dind/blob/master/Dockerfile#L22). They both works fine.

PR #165, opened 2020-02-13
Use bash as the default shell in Dockerfile

felipecassiors:
By doing this, we don't have to advertise people to use `RUN bash -c` anymore, such as [here](https://github.com/gitpod-io/website/blob/master/src/docs/java-in-gitpod.md#sdkman-in-docker).

Also, with bash as the default shell in `RUN` statements we can make use of `set -o pipefail` to enable safer `RUN` statements. More details [here](https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/). An example of implementation is these lines:

https://github.com/felipecassiors/jenkins-agent-dind/blob/2a77440b30c3c2b0471c4f59f2d63eb516f31c49/Dockerfile#L86-L90

In the example above, an error thrown in `wget` would prevent `tar` to run and assure the `RUN` statement returns a fail, because otherwise `RUN false | true` would result in a succeeded `RUN`, despite we probably want the opposite.


felipecassiors:
https://github.com/felipecassiors/jenkins-agent-dind/blob/2a77440b30c3c2b0471c4f59f2d63eb516f31c49/Dockerfile#L86-L90

felipecassiors:
I don't know why the code snippet with the permalink didn't work, but here is the code I mentioned:

```dockerfile
# Install shellcheck
RUN set -euxo pipefail; \
    wget -q 'https://storage.googleapis.com/shellcheck/shellcheck-stable.linux.x86_64.tar.xz' -O - \
    | tar -xJf - --strip-components=1 -C /usr/local/bin shellcheck-stable/shellcheck; \
    chmod +x /usr/local/bin/shellcheck
```

jankeromnes:
Excellent idea, many thanks! 💯

As mentioned in #166, I'll fix the CircleCI build so that we can fully test this improvement. I'm looking forward to it.

jankeromnes:
> I don't know why the code snippet with the permalink didn't work, but here is the code I mentioned:

Unfortunately, GitHub doesn't expand code permalinks from foreign repositories.

> In the example above, an error thrown in `wget` would prevent `tar` to run and assure the `RUN` statement returns a fail, because otherwise `RUN false | true` would result in a succeeded `RUN`, despite we probably want the opposite.

Interesting. But isn't the unwanted continuation caused by `RUN false; true` instead? I.e. I think that using `&&` instead of `;` would make the build fail without requiring `set -o pipefail`:

```Dockerfile
FROM buildpack-deps:disco

# Install shellcheck
RUN wget -q 'oops' -O - \
    | tar -xJf - --strip-components=1 -C /usr/local/bin shellcheck-stable/shellcheck && \
    chmod +x /usr/local/bin/shellcheck
```

fails with:

```
Sending build context to Docker daemon  2.048kB
Step 1/2 : FROM buildpack-deps:disco
 ---> 4f7094155bd0
Step 2/2 : RUN wget -q 'oops' -O -     | tar -xJf - --strip-components=1 -C /usr/local/bin shellcheck-stable/shellcheck &&     chmod +x /usr/local/bin/shellcheck
 ---> Running in 56536ca853d0
xz: (stdin): File format not recognized
tar: Child returned status 1
tar: Error is not recoverable: exiting now
The command '/bin/sh -c wget -q 'oops' -O -     | tar -xJf - --strip-components=1 -C /usr/local/bin shellcheck-stable/shellcheck &&     chmod +x /usr/local/bin/shellcheck' returned a non-zero code: 2
```

(well, it's actually the `tar` commands which throws the error, but having a `&&` instead of a `;` after it does make the Docker build fail as expected).

felipecassiors:
> > In the example above, an error thrown in `wget` would prevent `tar` to run and assure the `RUN` statement returns a fail, because otherwise `RUN false | true` would result in a succeeded `RUN`, despite we probably want the opposite.
> 
> Interesting. But isn't the unwanted continuation caused by `RUN false; true` instead? I.e. I think that using `&&` instead of `;` would make the build fail without requiring `set -o pipefail`:
> 
> ```dockerfile
> FROM buildpack-deps:disco
> 
> # Install shellcheck
> RUN wget -q 'oops' -O - \
>     | tar -xJf - --strip-components=1 -C /usr/local/bin shellcheck-stable/shellcheck && \
>     chmod +x /usr/local/bin/shellcheck
> ```
> 
> fails with:
> 
> ```
> Sending build context to Docker daemon  2.048kB
> Step 1/2 : FROM buildpack-deps:disco
>  ---> 4f7094155bd0
> Step 2/2 : RUN wget -q 'oops' -O -     | tar -xJf - --strip-components=1 -C /usr/local/bin shellcheck-stable/shellcheck &&     chmod +x /usr/local/bin/shellcheck
>  ---> Running in 56536ca853d0
> xz: (stdin): File format not recognized
> tar: Child returned status 1
> tar: Error is not recoverable: exiting now
> The command '/bin/sh -c wget -q 'oops' -O -     | tar -xJf - --strip-components=1 -C /usr/local/bin shellcheck-stable/shellcheck &&     chmod +x /usr/local/bin/shellcheck' returned a non-zero code: 2
> ```
> 
> (well, it's actually the `tar` commands which throws the error, but having a `&&` instead of a `;` after it does make the Docker build fail as expected).

## First

`set -e` before and using `;` to separate commands is equals to use `&&` to separate commands.

## Second

The new thing here is the `set -o pipefail`, which causes a pipe `|` to fail in the first failed command in the pipe (aka the most left failed one). Otherwise, it would only fail if the last command fails (the most at right).

Take a look: 
[asciicast](https://asciinema.org/a/vWNBerwvKZlHcGHE0VOZbxVL8)

In the example, you will see that `false | true && echo "final line"` will result in the last command to be still executed if not set `set o pipefail`, despite it is not the wanted situation.

felipecassiors:
A real and more interesting example using `RUN` statements and real commands (`wget` and `sh`).

[asciicast](https://asciinema.org/a/yvofdHfsb4mtIkccOTFsUuALz)

felipecassiors:
> Interesting. But isn't the unwanted continuation caused by `RUN false; true` instead? I.e. I think that using `&&` instead of `;` would make the build fail without requiring `set -o pipefail`:
> 
> ```dockerfile
> FROM buildpack-deps:disco
> 
> # Install shellcheck
> RUN wget -q 'oops' -O - \
>     | tar -xJf - --strip-components=1 -C /usr/local/bin shellcheck-stable/shellcheck && \
>     chmod +x /usr/local/bin/shellcheck
> ```
> 
> fails with:
> 
> ```
> Sending build context to Docker daemon  2.048kB
> Step 1/2 : FROM buildpack-deps:disco
>  ---> 4f7094155bd0
> Step 2/2 : RUN wget -q 'oops' -O -     | tar -xJf - --strip-components=1 -C /usr/local/bin shellcheck-stable/shellcheck &&     chmod +x /usr/local/bin/shellcheck
>  ---> Running in 56536ca853d0
> xz: (stdin): File format not recognized
> tar: Child returned status 1
> tar: Error is not recoverable: exiting now
> The command '/bin/sh -c wget -q 'oops' -O -     | tar -xJf - --strip-components=1 -C /usr/local/bin shellcheck-stable/shellcheck &&     chmod +x /usr/local/bin/shellcheck' returned a non-zero code: 2
> ```
> 
> (well, it's actually the `tar` commands which throws the error, but having a `&&` instead of a `;` after it does make the Docker build fail as expected).

Here, indeed it was the `tar` command which failed (the most at right one). But relying on the second command behavior may not be safe, as I showed in the previous comment with `wget` and `sh`.

felipecassiors:
Rebased, let's see the build results.

PR #164, opened 2020-02-12
[full] Fix gopls install

jankeromnes:
Fixes this build error:

```
package golang.org/x/tools/cmd/gopls: cannot find package "golang.org/x/tools/cmd/gopls" in any of:
  /home/gitpod/go/src/golang.org/x/tools/cmd/gopls (from $GOROOT)
  /home/gitpod/go-packages/src/golang.org/x/tools/cmd/gopls (from $GOPATH)
```

PR #166, opened 2020-02-13
Add support for Flutter web

felipecassiors:
Flutter now can run not only in Android and iOS, but on web. This PR enables a development environment for Flutter web. 

It can be tested here:
https://gitpod.io/#https://github.com/price-search/frontend

Preview: 
image

[Sample `.gitpod.yml`](https://github.com/price-search/frontend/blob/master/.gitpod.yml):
```yml
image:
  file: .gitpod.Dockerfile

tasks:
  - command: flutter run --web-port 8686

ports:
  - port: 8686
    onOpen: open-preview

vscode:
  extensions:
    - Dart-Code.flutter@3.8.1:cytqQi/+kpu802ZOaJwayw==
    - Dart-Code.dart-code@3.8.1:+5Rq24Mq46eJa5w39WY0bA==
```

## Why I renamed the old one

The Android based Gitpod Flutter environment will never work well: 
- Flutter requires a working Android emulator, something that can't be done in the Gitpod container since it doesn't have virtualization enabled. We could still use ARM images but they are very slow and already deprecated since a long time (the last one is Android 7.1).


svenefftinge:
This is cool, thank you! @jankeromnes could you review and check why the build fails?

jankeromnes:
Wow, super cool contribution, and works great! 💯 Many thanks.

The CircleCI build failed because of `gopls`, we're currently trying to fix that in https://github.com/gitpod-io/workspace-images/pull/164 (unrelated to this PR).

I'll finish fixing the `gopls` install, and also review this amazing PR. 👍

flutter/Dockerfile
```diff
@@ -1,25 +1,26 @@
-FROM gitpod/workspace-full-vnc
+FROM gitpod/workspace-full
 
+USER root
```
jankeromnes: Nit: In general, we now prefer using only `USER gitpod` everywhere, and using `sudo` where root-rights are required.

This prevents Gitpod users from accidentally creating directories as `USER root`, and then not having access to them in their workspaces (without an easy way to fix it).

flutter/Dockerfile
```diff
@@ -1,25 +1,26 @@
...
 
-ENV ANDROID_HOME=/home/gitpod/android-sdk-linux \
-    FLUTTER_HOME=/home/gitpod/flutter \
-    PATH=/usr/lib/dart/bin:$FLUTTER_HOME/bin:$ANDROID_HOME/tools:$PATH
+SHELL [ "/bin/bash", "-c" ]
```
jankeromnes: I see that this is already suggested in #165. I really like the idea of using `bash` instead of `sh`. Maybe we can merge this PR first, as a proof-of-concept, and then generalize this change to `gitpod/workspace-full`?

flutter/Dockerfile
```diff
@@ -1,25 +1,26 @@
...
 
-USER root
+### Install Flutter
+# dependencies
+RUN set -ex; \
+    apt-get update; \
+    apt-get install -y libglu1-mesa
```
jankeromnes: Our usual snippet for installing things with `apt-get` is:

```Dockerfile
USER gitpod
RUN sudo apt-get update && \
    sudo apt-get install -y libglu1-mesa && \
    sudo rm -rf /var/lib/apt/lists/*
```

I.e., user `gitpod` with `sudo` as mentioned above, and removing left-over data from `apt-get` to keep the image size small. This has historically worked well.

I'm not against changing this everywhere, but I'd like to understand your approach first. Specifically, what is the benefit of using `set -ex` and separating commands with `;` here?

jankeromnes:
@felipecassiors I've just fixed the CircleCI build in `master`. Could you please rebase this Pull Request, when you get a chance?

Or, if you don't have time, please let me know -- I'm happy to take over this PR. 🙂 Thanks again for making it!

flutter/Dockerfile
felipecassiors: That's right, let me change it.

flutter/Dockerfile
felipecassiors: This is completely up to you... however I will keep it here until https://github.com/gitpod-io/workspace-images/pull/165 don't get merged.

flutter/Dockerfile
felipecassiors: Done.

flutter/Dockerfile
felipecassiors: As in https://github.com/gitpod-io/workspace-images/pull/165#issuecomment-587039533, `set -e` with `;` behaves the same as `&&`. 

The main motivation comes from the following:

In my opinion we should always use `set -x` in any `RUN` statement with multi commands. This way, in a case where the build fails, we would know exactly what command failed in a set of them just looking in the logs. Much easier to debug. Take a look please:
[asciicast](https://asciinema.org/a/lh4qb5lEWQvWQh0aaAjbzj0Ys)

The second motivation is that `;` is the designed character to run multiple commands in a single line (which is how the multi-line in `RUN` is interpreted by the shell). ([reference](https://www.tldp.org/LDP/abs/html/special-chars.html)).

The third motivation is since we would use `set -x` already, adding a extra `e` character to it is easier than adding `&&` to every new command, considering that `;` is shorter than `&&`.

By the way [Dockerfile reference](https://docs.docker.com/engine/reference/builder/#run) uses `;`, despite they don't do any kind of error catch (`set -e` I mean).

felipecassiors:
> @felipecassiors I've just fixed the CircleCI build in `master`. Could you please rebase this Pull Request, when you get a chance?
> 
> Or, if you don't have time, please let me know -- I'm happy to take over this PR if you prefer. 🙂 Thanks again for making it!

Done. Let's see the results.

flutter/Dockerfile
felipecassiors: PS: I don't mind in settle in your current way. I just wanted to open the discussion for something I think is better.

jankeromnes: Thanks a lot for explaining your preference of `set -e` / `;` over `&&`!

It makes sense to me, and I'm happy to start using it in more places in this repo. 👍

jankeromnes:
Ah, I just realize that `flutter/Dockerfile` doesn't even get build on CircleCI yet. 🤦‍♂

It needs to be added in https://github.com/gitpod-io/workspace-images/blob/master/.circleci/config.yml -- I'll push a commit to this PR if you don't mind.

flutter/Dockerfile
felipecassiors: Oh, I thought I was using `bash -c` for some reason, but I am not. So I am removing this line.

felipecassiors:
Edited the commit messages to match your pattern and removed the `SHELL bash`.

jankeromnes:
FYI, the previous build succeeded (except for the two last jobs, but that seemed to be an intermittent CircleCI problem and not something related to your PR):

Screenshot 2020-02-17 at 19 03 28

jankeromnes:
Why did you remove the `SHELL bash` by the way?

jankeromnes:
Since the build of all images went through, and `gitpod/workspace-full` is a brand new image, I think we can merge this now without further testing.

After `master` deploy, we'll be able to test the new official `gitpod/workspace-flutter` image for web-based Flutter development. Many thanks @felipecassiors for this amazing contribution! 💯

PR #169, opened 2020-02-20
[full-vnc] Change the default number of Openbox virtual desktops from 4 to 1 (footgun)

jankeromnes:
Otherwise, when you scroll on a desktop, it constantly cycles between desktops.

That's super confusing and annoying.


jankeromnes:
Tested with https://github.com/jankeromnes/Hangman/commit/58241b11e7ad79b13da94fd0c72a255a97672b56, works as expected! 🎉

PR #170, opened 2020-02-20
[VNC] Use vnc.html

JesterOrNot:


Kreyren:
Recommends adding in-code documentation i.e

```html
<!-- 
Used to append options to vnc.html 
- autoconnect .. To automatically connect on the host
- resize=scale .. To scale the resolution based on used preview
-->
```

Kreyren:
Note that this solution has a bar on the left side of the screen 
image

Relevant: https://github.com/novnc/noVNC/issues/941

JesterOrNot:
How do we remove it?

full-vnc/novnc-index.html
```diff
@@ -1 +1 @@
-<html><head><meta http-equiv="Refresh" content="0; url=vnc_lite.html?scale=true"></head></html>
+<html><head><meta http-equiv="Refresh" content="0; url=vnc.html?scale=true&autoconnect=true"></head></html>
```
jankeromnes: I wonder what the difference between `?scale=true` and `?resize=scale` is.

Kreyren: true does not work on vnc.html on my end

JesterOrNot: Fixed

JesterOrNot:
We need to remove the sidebar how do we do this?

Kreyren:
> @JesterOrNot How do we remove it?

I am unable to find documentation for this.

Possibilities: 
1. https://github.com/novnc/noVNC/pull/725/files ? May be possible to manipulate ` app/styles/base.css`
2. wip

jankeromnes:
> We need to remove the sidebar how do we do this?

I think we want to keep it. I've found it useful before, especially on mobile (e.g. for copy/pasting and such).

JesterOrNot:
Can someone test this PR?

Kreyren:
LGTM

JesterOrNot:
Merge @jankeromnes?

Kreyren:
Seems that CI hugged up
image

meysholdt:
Will this PR also add 
image when the VNC client runs in the Preview View in Gitpod? I'm a bit worried there it adds more clutter than it adds value. It would distract people from working with their apps and damages the feeling that things are "simple". What use cases does it solve?

JesterOrNot:
@meysholdt It would :(

Kreyren:
@meysholdt referencing https://github.com/gitpod-io/workspace-images/pull/170#issuecomment-589103465

I guess that asking upstream again is the only option to get more info based on the fact that we don't have documentation on this

jankeromnes:
> Can someone test this PR?

I've tested it by opening https://gitpod.io/#https://github.com/techwithtim/Hangman and running this in the Terminal:

```
gp preview "$(gp url 6080)/vnc.html?resize=scale&autoconnect=true"
```

Works like a charm, and I don't find the controls too distracting. Cool that users can now benefit from these by default. 🙂

> EDIT: i've would add the in-code documentation

I agree that documentation could be nice, e.g. in cases like these where it's unclear what some things mean or do exactly.

But it's not a strong requirement on my part, so I'll let @JesterOrNot make the call of whether to add the comment, or keep this a short-and-sweet HTML one-liner. 🙂

JesterOrNot:
@Kreyren We could always file another issue

Kreyren:
@JesterOrNot your turn, be nice! :p

Kreyren:
> I've tested it by opening https://gitpod.io/#https://github.com/techwithtim/Hangman and running this in the Terminal:
>
> gp preview "$(gp url 6080)/vnc.html?resize=scale&autoconnect=true"

Would be also nice to have `README.md` in the vnc-full directory that informs about this in case anyone wants to work on vnc in the future to know how to properly test this.

JesterOrNot:
I decided to add the comments because they can only help :)

Kreyren:
Also side-note: 
image

In theory we can customize this to fix https://github.com/gitpod-io/gitpod/issues/1193

(should be issues as separate issue)

meysholdt:
Tied it, too:
* the screen is unusable because of the strong zoom factor. There should be no zoom at all. The virtual screen should have the resolution of the preview widget.
* the "fullscreen" button is broken with a JavaScript error `TypeError: fullscreen error
    at HTMLInputElement.toggleFullscreen (https://1233234234.ws-eu01.gitpod.io/app/ui.js:1215:42)`

Kreyren:
> the screen is unusable because of the strong zoom factor. There should be no zoom at all. The virtual screen should have the resolution of the preview widget.

image

Works for me once it is on appropriate viewing screen.

in theory we can use `xrandr --output OUTPUT --scale 2x2` for this while making some kind of GUI that would allow end-user to change the scaling on demand.

> the "fullscreen" button is broken with a JavaScript error TypeError: fullscreen error at HTMLInputElement.toggleFullscreen (https://1233234234.ws-eu01.gitpod.io/app/ui.js:1215:42)

WFM

Kreyren:
EDIT: 

I guess that the font size is an issue?

image

jankeromnes:
> the screen is unusable because of the strong zoom factor. There should be no zoom at all. The virtual screen should have the resolution of the preview widget.

When I try `resize=downscale` or `resize=remote` instead of `resize=scale`, it doesn't seem to do anything. Maybe this should be a bug report or feature request to noVNC?

> the "fullscreen" button is broken with a JavaScript error TypeError: fullscreen error at HTMLInputElement.toggleFullscreen (https://1233234234.ws-eu01.gitpod.io/app/ui.js:1215:42)

Maybe that's because iframes are not allowed to go fullscreen in some browsers? (E.g. to prevent annoying fullscreen ads taking over your screen).

Kreyren:
> Maybe that's because iframes are not allowed to go fullscreen in some browsers? (E.g. to prevent annoying fullscreen ads taking over your screen).

UserAgent: Mozilla/5.0 (X11; Linux x86_64; rv:68.0) Gecko/20100101 Firefox/68.0

(don't know how else to sent browser info)
image

Kreyren:
> For me, fullscreen button shows the error when in a preview pane, but it works as expected when the preview is opened in a new tab.

Sorry did not test it in gitpod, same issue here.

Kreyren:
What is the proposed course of action for this? i've check the available documentation and i can't find anything relevant to resolve the scaling issue atm.

JesterOrNot:
@Kreyren Also can you talk to the noVNC people again I don't know what you meant https://github.com/novnc/noVNC/issues/1371

Kreyren:
@JesterOrNot ehh what

JesterOrNot:
> @JesterOrNot your turn, be nice! :p



> Me: @Kreyren We could always file another issue
> You: @JesterOrNot your turn, be nice! :p

Kreyren:
@JesterOrNot Figured that much, but what do you expect from me assuming that it seems that you've described the issue sufficiently 

also why is making a button in GUI that would changed the css style not an option? Or how do you expect it to be toggleable ?

JesterOrNot:
> This is a commonly discussed when integrating noVNC in to other systems, so I think this is worth having open and investigating further.
> 
> @JesterOrNot, the toolbar is needed for many user interactions (e.g. touch, clipboard). Is this something you've considered? Perhaps there are other ways than just completely removing the toolbar.

Kreyren:
iiuc the clipboard and touch capabilities are the concern here?

For clipboard i guess that the proper implementation would be use the end-user's clipboard if browser allows it.

For touch i don't think it's worth processing until gitpod has better implementation for portable devices (https://github.com/gitpod-io/gitpod/issues/970) assuming that none would probably want to use touch on PC version (personally i don't see the usecase for it there)

JesterOrNot:
What does iiuc mean?

Kreyren:
iiuc == if i understand correctly

added in https://en.wikipedia.org/w/index.php?title=SMS_language&action=history

jankeromnes:
For the resolution, I think we need to fix this in the X server config.

E.g. we could divide the Xvfb screen size by 2, maybe this will help make things more legible?

jankeromnes:
Ok, I'm going to clean up and merge this PR. Thanks again!

I've also enabled `reconnect=true` (`autoconnect=true` was a good idea, but that actually just skips the initial screen with "Connect" button -- to actually reconnect when the connection breaks, you need `reconnect=true`)

FYI, here are all the `vnc.html` options: https://github.com/novnc/noVNC/blob/master/docs/EMBEDDING.md

Also, having the noVNC preview automatically reconnect on connection breaks is compelling enough to switch from `vnc_lite.html` to `vnc.html`.

For what it's worth, I find the extra controls very useful, but if someone else really doesn't like them, we could potentially contribute a `hidecontrols=true` option to noVNC upstream.

Issue #171, opened 2020-02-21
Postgres images: persist data after workspace stop


marcopeg:
Hello, I'm running the postgres image and all works fine until I stop/start the workspace and all the data are lost.

I read that data under `/workspace` are persisted, but I can't find a way to move the postgres data folder in there when extending the Dockefile:  
https://github.com/marcopeg/hasura-gitpod/blob/master/.gitpod.Dockerfile

Can you help?

jankeromnes:
Hi @marcopeg, many thanks for reporting this issue!

Thanks to @jjangga0214, this problem was fixed in https://github.com/gitpod-io/workspace-images/pull/172 and an updated `gitpod/workspace-postgresql` default image will be available on Docker Hub and in Gitpod shortly (ETA about ~1 hour).

We've changed the default location for PostgreSQL data to `/workspace/.pgsql/data`, which would solve your issue without requiring any change.

(Still, if you desire to change the location, `$PGDATA` is now configurable via https://gitpod.io/environment-variables/, but in your case that's no longer necessary.)

PR #172, opened 2020-02-22
fix(postgres): use $PGDATA

jjangga0214:
There're 2 problems.

1.  $PGDATA is only declared, but not used.
2. User cannot configure $PGDATA (for example, they might want it to be a dir in /workspace, like #171) 

This tries to resolve them. 

Note that this is not tested, so reviewers should do it, if CI is not enough.

P.S. `printf "..."` and `printf '...'` are different. env var would be interpolated with "", but not with ''. For instance, try `printf "$foo"` and `printf '$foo'` in your local machine.

### DIFF
--- a/postgres/Dockerfile
+++ b/postgres/Dockerfile
@@ -9,10 +9,9 @@ RUN sudo apt-get update \
 # Setup PostgreSQL server for user gitpod
 ENV PATH="$PATH:/usr/lib/postgresql/11/bin"
 ENV PGDATA="/home/gitpod/.pg_ctl/data"
-RUN mkdir -p ~/.pg_ctl/bin ~/.pg_ctl/data ~/.pg_ctl/sockets \
- && initdb -D ~/.pg_ctl/data/ \
- && printf "#!/bin/bash\npg_ctl -D ~/.pg_ctl/data/ -l ~/.pg_ctl/log -o \"-k ~/.pg_ctl/sockets\" start\n" > ~/.pg_ctl/bin/pg_start \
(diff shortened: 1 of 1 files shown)


postgres/Dockerfile
```diff
@@ -23,3 +22,4 @@ ENV PGDATABASE="postgres"
 # tasks from a Dockerfile. This workaround checks, on each bashrc eval, if the
 # PostgreSQL server is running, and if not starts it.
 RUN printf "\n# Auto-start PostgreSQL server.\n[[ \$(pg_ctl status | grep PID) ]] || pg_start > /dev/null\n" >> ~/.bashrc
+CMD ["sh", "-c", "'mkdir -p $PGDATA && initdb -D $PGDATA'", "&&", "sh"]
```
jankeromnes: Unfortunately, setting `CMD` won't work, because Gitpod needs it and overwrites it just before starting a workspace.

Instead, I think we should use `~/.bashrc` to `mkdir` and `initdb` on-the-fly when needed. (Or maybe `pg_start` should do it?)

jankeromnes: I took the liberty of fixing this myself by pushing a commit to your PR. 🙂 I hope you don't mind.

jankeromnes:
Notes:
- I moved the `$PGDATA` initialization from `CMD` to `pg_start`
- I made `PGDATA` be `/workspace/.pgsql/data` by default (so persistence in Gitpod workspaces should work by default, no need for extra configuration, although that remains possible too)

marcopeg:
:+1: thanks guys

PR #174, opened 2020-02-28
Update Dockerfile to install both 2.2 and 3.1

orellabac:
I am including both .NET Core 2.2 and .Net Core 3.1 because I have found that some `dotnet tools` are still on .netcore2.2

### DIFF
--- a/dotnet-vnc/Dockerfile
+++ b/dotnet-vnc/Dockerfile
@@ -6,4 +6,5 @@ RUN wget -q https://packages.microsoft.com/config/ubuntu/18.04/packages-microsof
     dpkg -i packages-microsoft-prod.deb && rm -rf packages-microsoft-prod.deb && \
     add-apt-repository universe && \
     apt-get update && apt-get -y -o APT::Install-Suggests="true" install dotnet-sdk-2.2 && \
-    apt -y clean;
+    apt-get update && apt-get -y -o APT::Install-Suggests="true" install dotnet-sdk-3.1 && \
+    rm -rf /var/lib/apt/lists/*;


jankeromnes:
Hi @orellabac, many thanks for this contribution!

It seems like I forgot about the `gitpod/workspace-dotnet-vnc` image when I upgraded `gitpod/workspace-dotnet` to .NET Core 3.1 in 3f81f46b3a46721cd39739dcff427ec76d637bd4... 🙈

However, I'm curious: Why would you like to keep .NET Core 2.2 also installed? Which `dotnet tools` are you referring to that still use it?

orellabac:
Hi @jankeromnes 
For example a lot of the tools in:
https://github.com/natemcmaster/dotnet-tools
Have not yet being ported to 3.1 so if your build process uses them then it will not work.
In my case i use several of them even some of my own which I had not yet ported to 3.1

jankeromnes:
> For example a lot of the tools in:
> https://github.com/natemcmaster/dotnet-tools
> Have not yet being ported to 3.1 so if your build process uses them then it will not work.
> In my case i use several of them even some of my own which I had not yet ported to 3.1

Good point, many thanks @orellabac. Then I think both official Gitpod .NET images should have 2.2 and 3.1, i.e.:
- `gitpod/workspace-dotnet-vnc` (the one you've fixed here)
- `gitpod/workspace-dotnet`

jankeromnes:
Update: I took the liberty of cherry-picking your commit into #175, because we were also modifying the .NET Dockerfiles over there.

Many thanks again for your valuable contribution! And also for the insight on 2.2 vs 3.1, which is really precious to us, as we're not .NET developers ourselves.

orellabac:
Thanks to you. Gitpod is a great tool

PR #179, opened 2020-03-09
[full] Update python version

JesterOrNot:
Update Python version from 3.7.6 to 3.8.2

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -207,8 +207,8 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
         echo 'eval "$(pyenv init -)"'; \
         echo 'eval "$(pyenv virtualenv-init -)"'; } >> /home/gitpod/.bashrc.d/60-python \
     && pyenv install 2.7.17 \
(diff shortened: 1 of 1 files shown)


jankeromnes:
Thanks @JesterOrNot! See also https://github.com/gitpod-io/workspace-images/pull/145#pullrequestreview-339147832

jankeromnes:
@RDIL We'd like to upgrade to Python `3.8.2`. Do you think that it still might break projects like TensorFlow? Or is it reasonable to expect that they should work with `3.8` by now?

RDIL:
let me check

JesterOrNot:
Before merging, we should also update the Python language guide

RDIL:
Tensorflow is still constrained, so this can't be merged.

jankeromnes:
Noted, thanks a lot!

If TensorFlow remains constrained for a long time, we may want to look into having a TensorFlow specific config that pins `3.7` while upgrading the rest of the world to `3.8+`.

RDIL:
Please merge #182 in the meantime since it still updates python, and #152

aledbf:
Closing. Outdated.

PR #181, opened 2020-03-11
Increase no_output timeout

JesterOrNot:
Should fix workspace-dotnet build

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,7 +70,7 @@ jobs:
           command: |
             docker load -i workspace-full.tar
             ./.circleci/build_image.sh dotnet/Dockerfile gitpod/workspace-dotnet
-          no_output_timeout: 45m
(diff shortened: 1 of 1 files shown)


JesterOrNot:
@jankeromnes So adjust the other values to 45m?

jankeromnes:
@JesterOrNot Or better, imitate the status-quo by using values that already exist.

FYI,
- `10m` is the CircleCI default (so when it's not specific it's 10 minutes)
- some jobs got bumped to `30m` because 10 wasn't enough
- other jobs got bumped to `60m`

In general, I feel like when you need to raise a limit, you should double it. This avoids having to constantly raise limits all the time. 😄 (I think I got that from Google engineers.)

JesterOrNot:
better?

jankeromnes:
Awesome, many thanks! 🎉 (And sorry for complaining about such little things!)

PR #186, opened 2020-03-23
Install amazing tools

### LINKED ISSUE #76: Pre-install popular/amazing CLI tools
E.g.

- [x] `jq`
- [ ] `rg`
- [ ] `fd`
- [ ] `z`
- [ ] `fzf`
- [ ] `mtr`
- [ ] `hub` (#77)

Inspiration: [this thread](https://twitter.com/krisnova/status/1109618657305333761) and [this post](https://remysharp.com/2018/08/23/cli-improved)

Also maybe add tips about them in the Gitpod loading screen, see https://github.com/gitpod-io/gitpod/issues/93

JesterOrNot:
Fixes #76

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -8,6 +8,15 @@ RUN yes | unminimize \
         build-essential \
         htop \
         jq \
+        hub=2.7.0~ds1-1 \
+        zsh=5.5.1-1ubuntu3 \
+        tmux=2.8-3 \
+        fzf=0.17.5-2 \
+        mtr=0.92-2 \
+        ripgrep=0.10.0-2 \
+        neovim=0.3.4-1ubuntu0.19.04.1 \
+        fd-find=7.2.0-2 \
+        silversearcher-ag=2.2.0-1 \
         less \
         locales \
         man-db \
@@ -19,7 +28,10 @@ RUN yes | unminimize \
         lsof \
     && locale-gen en_US.UTF-8 \
     && mkdir /var/lib/apt/dazzle-marks \
-    && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/*
+    && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* \
+    && wget "https://raw.githubusercontent.com/rupa/z/master/z.sh" \
+    && mv z.sh /home/z.sh \
+    && echo -e "\n. /home/z.sh" >> $HOME/.bashrc
 
 ENV LANG=en_US.UTF-8
 
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -149,6 +149,7 @@ RUN curl -fsSL https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.t
         github.com/josharian/impl \
         github.com/tylerb/gotype-live \
         github.com/rogpeppe/godef \
+        github.com/jesseduffield/lazygit \
         github.com/zmb3/gogetdoc \
         golang.org/x/tools/cmd/goimports \
         github.com/sqs/goreturns \


full/Dockerfile
```diff
@@ -8,6 +8,15 @@ RUN yes | unminimize \
         build-essential \
         htop \
         jq \
+        hub=2.7.0~ds1-1 \
+        zsh=5.5.1-1ubuntu3 \
+        tmux=2.8-3 \
+        fzf=0.17.5-2 \
+        mtr=0.92-2 \
+        ripgrep=0.10.0-2 \
+        neovim=0.3.4-1ubuntu0.19.04.1 \
```
RDIL: We have theia

full/Dockerfile
```diff
@@ -8,6 +8,15 @@ RUN yes | unminimize \
         build-essential \
         htop \
         jq \
+        hub=2.7.0~ds1-1 \
```
RDIL: Just use git

full/Dockerfile
```diff
@@ -8,6 +8,15 @@ RUN yes | unminimize \
...
+        zsh=5.5.1-1ubuntu3 \
```
RDIL: We have bash

full/Dockerfile
```diff
@@ -8,6 +8,15 @@ RUN yes | unminimize \
...
+        tmux=2.8-3 \
+        fzf=0.17.5-2 \
+        mtr=0.92-2 \
+        ripgrep=0.10.0-2 \
```
RDIL: We have grep

full/Dockerfile
JesterOrNot: See #76

full/Dockerfile
JesterOrNot: see #76

full/Dockerfile
JesterOrNot: See #76

full/Dockerfile
JesterOrNot: See #76

jankeromnes:
Many thanks @RDIL for ensuring that our default image doesn't get too big! 💯 I agree that this is important.

full/Dockerfile
jankeromnes: Personally, I like `hub`, because it allows me to run `hub pull-request` (something that `git` doesn't directly allow, and that usually requires jumping around various web pages to achieve).

But if we do want to pre-install `hub` in all Gitpod workspaces, it should also work perfectly out-of-the box.

Have you tried to `brew install hub` in a running Gitpod workspace, and did it work well? (I suspect that it requires some extra configuration before it can work, e.g. setting up a GitHub Personal Access Token.)

full/Dockerfile
jankeromnes: I've used `bash` all my life, but recently while helping out @carl-parrish to get `zsh` working in Gitpod, I realized how nice `zsh` is to use. I can't really explain why, but the experience seemed surprisingly smooth and fast.

Long story short, I don't mind having `zsh` pre-installed in Gitpod (provided that it doesn't make the default image too large), and maybe it can eventually even become the default shell if everyone likes it. (But we'd probably need to have a sort of transition phase where people can easily switch between `bash` and `zsh`, in order not to badly surprise users.)

full/Dockerfile
jankeromnes: We already use `rg` in Gitpod (it's the back-end for the workspace search -- that's why it's so mind-blowingly fast).

But it would be cool to have `rg` accessible from the Terminal as well. Maybe we should pay attention not to install it twice (unless its installed size is negligible).

full/Dockerfile
jankeromnes: I like the idea of having both `vim` and `neovim` in our images. It's just an extra quality-of-life thing for developers who usually use `neovim`, and discover it's available in Gitpod without extra effort.

(But again, only if its installed size is insignificant. If it's significant, we'd have to really evaluate usefulness vs cost.)

full/Dockerfile
```diff
@@ -8,6 +8,15 @@ RUN yes | unminimize \
...
+        neovim=0.3.4-1ubuntu0.19.04.1 \
+        fd-find=7.2.0-2 \
+        silversearcher-ag=2.2.0-1 \
```
jankeromnes: Maybe we should first decide on a subset in the suggested tools that we estimate are _extremely_ useful?

I don't want to install tons of (sometimes duplicate) tools, which then turn out not to get much usage. 😅

RDIL: Right.  I also feel like we need to evaluate the size of each language, most likely dropping off a ruby version.

full/Dockerfile
nicolo-ribaudo: I think `hub` should be replaced with `gh`: https://github.com/cli/cli

JesterOrNot:
First off let's close this PR and evaluate the tools in their own issues so we can debate merits and size costs on a case by case basis

JesterOrNot:
I created individual Issues for 3 of the primary tools so that talks on each tool's merits can be organized.

PR #190, opened 2020-03-25
Update python language server

### LINKED ISSUE #189: Time to upgrade the Python language server
While I might hate Palantir, I do think we need to update the PLS.  It is pinned on 0.25.0, which is roughly 2 years out-of-date.  

cc @jankeromnes @JesterOrNot as per usual 😄 

EDIT: Pull request now open!

RDIL:
2 years of commits being pulled in :)

Fixes #189


RDIL:
Doing this in another pr

PR #191, opened 2020-03-31
Prepend Brew's paths to env vars

Lctrs:
So they take precedence over pre-installed packages.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -110,9 +110,9 @@ LABEL dazzle/layer=tool-brew
 LABEL dazzle/test=tests/tool-brew.yaml
 USER gitpod
 RUN mkdir ~/.cache && sh -c "$(curl -fsSL https://raw.githubusercontent.com/Linuxbrew/install/master/install.sh)"
-ENV PATH="$PATH:/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin/" \
-    MANPATH="$MANPATH:/home/linuxbrew/.linuxbrew/share/man" \
-    INFOPATH="$INFOPATH:/home/linuxbrew/.linuxbrew/share/info" \
+ENV PATH="/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:$PATH" \
+    MANPATH="/home/linuxbrew/.linuxbrew/share/man:$MANPATH" \
+    INFOPATH="/home/linuxbrew/.linuxbrew/share/info:$INFOPATH" \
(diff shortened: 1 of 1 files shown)


Lctrs:
Following https://github.com/Orbitale/Compotes/pull/3

This way, when we install PHP 7.4 through brew, it takes precedence over pre-installed PHP 7.2.

Friendly ping @jankeromnes

jankeromnes:
Hi @Lctrs, many thanks for suggesting this change!

We originally had all HomeBrew `PATH`s prefixed like that, and not suffixed, but I remember that it was breaking something with OpenSSL, which is why we've decided to go with the suffix.

Another way to let HomeBrew control the PHP version would be to uninstall the `apt-get` PHP before installing it via `brew`. Then you wouldn't have any `PATH` problem I think.

But still, I'm a little bit worried about image size, which is why I suggested sticking with `apt-get`. Would be interesting to compare the footprint of `brew install php` vs `apt-get install php && rm sources`. If it's comparable, then I'm definitely on board with managing PHP via HomeBrew and not apt.

jankeromnes:
(Sorry, didn't mean to close -- I clicked on the wrong button 🤦‍♂)

jankeromnes:
We could however wait for CircleCI to finish building the images, and check if the OpenSSL problem still happens with prefixed PATHs.

Lctrs:
I like how Java's versions are currently managed through SDKMAN!. I just thinks that PHP versions could be managed in a similar way through HomeBrew. Have a pre-installed version and if you need another one, just install it through brew and it just works.

jankeromnes:
> I like how Java's versions are currently managed through SDKMAN!. I just thinks that PHP versions could be managed in a similar way through HomeBrew. Have a pre-installed version and if you need another one, just install it through brew and it just works.

Thanks, that's a good point.

In your experience, do PHP developers often need a specific PHP version that is not the latest stable? E.g. I agree that for Java it makes sense to have a version manager, since so many projects have dependencies to different specific versions, but for other languages (e.g. Rust or Go) developers seem generally happy with the latest stable version, and so the complexity of a version manager can be avoided.

(Then again, if the version manager is `brew`, that's already installed by default, so I agree that we might as well use it for PHP too -- provided PHP's installed size is not enormous when installed with `brew` as opposed to `apt-get`.)

Lctrs:
We don't often need a specific patch version - unless we hit a bug in PHP, which can happen but is not very common I think.

But we do switch between minor versions very often.

Lctrs:
Using [asdf](https://asdf-vm.com) could also be an option to manage PHP version.

PR #194, opened 2020-04-02
Refactor of workspace-full

RDIL:
Haven't checked the exact savings yet, but I guarantee it will be at the very least 100 mb.

Changes:
* Rely on system python2.  Anybody still using those versions should know they are on their own now 👀
* Update python language server - should have better performance
* Remove ruby 2.5 - users can add it themselves if they really need it.  this should be the thing helping the size.  
* Build python 3 from source **!!!!!!!**
  * Pyenv breaks stuff so pyenv bad
  * pyenv gone now
* Implements #146 

NOTE: 1 change made is that now, full optimizations are being made to the python install.  This makes the build super slow but **gives an up to 25% performance boost for python**!

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -223,7 +223,6 @@ RUN sudo apt-get update \
         --with-ensurepip=install \
     && sudo make -j 8 \
     && sudo make install \
-    && sudo chown gitpod:gitpod /home/gitpod/.cache \
     && sudo chown gitpod:gitpod /home/gitpod/.local/lib/python${GP_PYTHON_SHORT_VERSION}/site-packages \
     && sudo ldconfig \
     && cd .. \
@@ -235,6 +234,7 @@ RUN sudo apt-get update \
         setuptools wheel virtualenv pipenv pylint rope flake8 \
         mypy autopep8 pep8 pylama pydocstyle bandit notebook \
         python-language-server[all]==0.31.9 twine \
+    && sudo chown gitpod:gitpod /home/gitpod/.cache \
     && curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash \
     && { echo; \
         echo 'eval "$(pyenv init -)"'; \
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -217,7 +217,7 @@ RUN sudo apt-get update \
     && sudo rm -rf Python-${GP_PYTHON_VERSION} Python-${GP_PYTHON_VERSION}.tar.xz \
     && curl -fsSL https://bootstrap.pypa.io/get-pip.py -o get-pip.py \
     && python get-pip.py \
-    && rm -rf get-pip.py \
+    && rm get-pip.py \
     && python3 -m pip install --upgrade pip \
     && python3 -m pip install --upgrade \
         setuptools wheel virtualenv pipenv pylint rope flake8 \
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -233,6 +233,7 @@ RUN sudo apt-get update \
         setuptools wheel virtualenv pipenv pylint rope flake8 \
         mypy autopep8 pep8 pylama pydocstyle bandit notebook \
         python-language-server[all]==0.31.9 twine \
+    && sudo mkdir /home/gitpod/.cache \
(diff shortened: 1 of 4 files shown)


RDIL:
cc @JesterOrNot for initial review

full/Dockerfile
```diff
@@ -204,41 +204,41 @@ ENV PATH=$PATH:/home/gitpod/.nvm/versions/node/v${NODE_VERSION}/bin
...
+    && tar xsf Python-${GP_PYTHON_VERSION}.tar.xz \
+    && cd Python-${GP_PYTHON_VERSION} \
+    && ./configure --enable-optimizations --enable-shared --with-lto --enable-ipv6 --enable-loadable-sqlite-extensions --without-ensurepip \
+    && sudo make -j 8 \
+    && sudo make install \
+    && cd .. \
+    && sudo rm -rf Python-${GP_PYTHON_VERSION} Python-${GP_PYTHON_VERSION}.tar.xz \
+    && curl -fsSL https://bootstrap.pypa.io/get-pip.py -o get-pip.py \
+    && python get-pip.py \
+    && rm -rf get-pip.py \
```
JesterOrNot, line 220: why is this rm recursive

full/Dockerfile
```diff
@@ -204,41 +204,41 @@ ENV PATH=$PATH:/home/gitpod/.nvm/versions/node/v${NODE_VERSION}/bin
 LABEL dazzle/layer=lang-python
 LABEL dazzle/test=tests/lang-python.yaml
 USER gitpod
-ENV PATH=$HOME/.pyenv/bin:$HOME/.pyenv/shims:$PATH
-RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash \
```
JesterOrNot, line 208: I don't think we should remove pyenv, it makes things easier to gitpodify and removing might break a *lot* of builds.

full/Dockerfile
RDIL: because why not

full/Dockerfile
RDIL: pyenv basically breaks a ton of tools, such as poetry, and doesn't work well in my opinion

full/Dockerfile
RDIL, line 220:
```suggestion
    && rm get-pip.py \
```

full/Dockerfile
JesterOrNot: How do you suggest people who do not know docker manage python versions in gitpod?

RDIL: Using python as the base docker image in the gp yml?

I'm going to add back Pyenv, but am still going to install the main python version like this.

The reason this is better is that it literally overrides the system preinstalled version of python, as well as fixes a lot of issues.

JesterOrNot: Good plan!

full/Dockerfile
```diff
@@ -204,41 +204,57 @@ ENV PATH=$PATH:/home/gitpod/.nvm/versions/node/v${NODE_VERSION}/bin
...
+    GP_PYTHON_SHORT_VERSION=3.7 \
+    PATH=$HOME/.pyenv/bin:$HOME/.pyenv/shims:/home/gitpod/.local/bin:$PATH
+
+RUN sudo apt-get update \
+    && sudo apt-get install -yq libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev llvm libncurses5-dev libncursesw5-dev xz-utils \
+    && curl -fsSL https://www.python.org/ftp/python/${GP_PYTHON_VERSION}/Python-${GP_PYTHON_VERSION}.tar.xz -o Python-${GP_PYTHON_VERSION}.tar.xz \
+    && tar xsf Python-${GP_PYTHON_VERSION}.tar.xz \
+    && cd Python-${GP_PYTHON_VERSION} \
+    && ./configure \
+        #--enable-optimizations \
```
RDIL, line 218: Note: this needs to be made before this is merged, **and is very much worth it but slows down the build a ton**.

```suggestion
        --enable-optimizations \
```

RDIL:
@JesterOrNot please rereview

RDIL:
Need to do this in another PR for some reason

Issue #193, opened 2020-04-02
FSharp interactive not properly installed


simonelnahas:
Running the command gives an error in gitpod, but it works when running locally.

```
gitpod /workspace/FsLexYaccServerTemplate $ fsharpi analyser/program.fsx 


error FS0193: Specified method is not supported.
```
The error should be replicable by opening this [commit](https://github.com/simonelnahas/FsLexYaccServerTemplate/tree/add50d3c4c81f8e7ffd7fb79d07d199733c3a046) on GitPod.

simonelnahas:
The `$ fsharpi ` command works, but this issue suggests there is something wrong with the installation:
https://github.com/fsharp/fsharp/issues/740

simonelnahas:
may do something similar to this: https://github.com/nblockchain/geewallet/commit/0326aa3564edd4c5a3d14ae6c09c0dbde3253f48

svenefftinge:
Hey Sean (@JesterOrNot), could you have a look into this? It seems like the mono version of `fsharpi` doesn't work with our version of ubuntu (see the issue referenced above). We probably should use an image that is not based on our workspace-full for this.

JesterOrNot:
Hey, sorry this hasn't worked for you :( I will continue to look into this. In the meantime I have fixed some errors in your .gitpod.yml.  https://github.com/simonelnahas/FsLexYaccServerTemplate/pull/2

JesterOrNot:
Hello Almost done I have it working now with the test branch!

![image](https://user-images.githubusercontent.com/50386709/78849410-0744c400-79da-11ea-82b2-a65d06deabb8.png)

Now all that should be left is optimization!

simonelnahas:
Cool i'm looking forward to it! :)

jankeromnes:
@JesterOrNot What is the status of this issue? Are there still some changes required to Gitpod's `dotnet` Dockerfiles, or is FSharp interactive now properly installed?

sr229:
Just a note, this is now obsolete as dotnet tools now handle F# within .NET 5.0. run your F# tools with `dotnet fsi` instead.

@jankeromnes I suggest closing this now since this is more of a fixed issue now :D

princerachit:
Closing this PR as the issue is fixed in recent versions of dotnet

PR #195, opened 2020-04-07
Fix fsharpi

JesterOrNot:
Fixes gitpod-io/workspace-images#193


dotnet/Dockerfile
```diff
@@ -1,15 +1,22 @@
...
     rm -rf packages-microsoft-prod.deb && \
     add-apt-repository universe && \
     apt-get update && \
-    echo resolvconf resolvconf/linkify-resolvconf boolean false | debconf-set-selections && \
     DEBIAN_FRONTEND=noninteractive ACCEPT_EULA=Y apt-get -y -o APT::Install-Suggests="true" install \
         dotnet-sdk-2.2 \
         dotnet-sdk-3.1 \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN apt-key adv --keyserver "hkp://keyserver.ubuntu.com:80" --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF \
```
RDIL, line 15: more layers bad

remove useless layer

JesterOrNot: Done!

dotnet/Dockerfile
```diff
@@ -1,15 +1,20 @@
...
-RUN wget -q https://packages.microsoft.com/config/ubuntu/19.04/packages-microsoft-prod.deb && \
+
+RUN wget -q "https://packages.microsoft.com/config/ubuntu/19.04/packages-microsoft-prod.deb" && \
     dpkg -i packages-microsoft-prod.deb && \
     rm -rf packages-microsoft-prod.deb && \
     add-apt-repository universe && \
     apt-get update && \
-    echo resolvconf resolvconf/linkify-resolvconf boolean false | debconf-set-selections && \
-    DEBIAN_FRONTEND=noninteractive ACCEPT_EULA=Y apt-get -y -o APT::Install-Suggests="true" install \
+    DEBIAN_FRONTEND=noninteractive ACCEPT_EULA=Y apt-get --no-install-recommends -y -o APT::Install-Suggests="true" install \
```
jankeromnes, line 10: In my experience, `--no-install-recommends` will fail to install packages that are expected by downstream developers and programs.

For example, this was the reason that https://github.com/MovingBlocks/Terasology mysteriously refused to work in Gitpod. It only started working again when we removed `--no-install-recommends`. So, I don't like this option (sure, it saves a few bytes in the image, but is that really worth it if developers then spend hours/days struggling to get their applications working?)

dotnet/Dockerfile
```diff
@@ -1,15 +1,20 @@
...
         dotnet-sdk-2.2 \
         dotnet-sdk-3.1 \
+    && apt-key adv --keyserver "hkp://keyserver.ubuntu.com:80" --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF \
+    && echo "deb https://download.mono-project.com/repo/ubuntu stable-bionic main" | tee /etc/apt/sources.list.d/mono-official-stable.list \
+    && apt-get update \ 
+    && apt-get install -yq \
```
jankeromnes, line 16: It feels a bit strange to me to do `apt-get update` and `apt-get install` twice in a row.

Do you think it's possible to do all the `add-apt-repository` and `apt-key adv` and `tee /etc/apt/sources.list` _before_, and then run `apt-get update` once, and then `apt-get install` all the things in one go?

jankeromnes: Also I think it's better to not use `USER root`, as that encourages others to do dangerous things that could prevent them from being able to access their own files.

Better to use `USER gitpod` (which is the default) and then use `sudo` everywhere it's needed.

RDIL: I agree with Jan, permission issues are a huge pain to deal with.

dotnet/Dockerfile
JesterOrNot: Done!

dotnet/Dockerfile
JesterOrNot: Done! And Stable!

simonelnahas:
if i run the image you made on my gitpod it still won't boot.
https://github.com/simonelnahas/FsLexYaccServerTemplate/blob/gpWorkspace/.gitpod.Dockerfile

JesterOrNot:
@jankeromnes could we look into reverting to the previous commit that works apparently the steps need to be done after.

simonelnahas:
I can't get neither the current or the previous commit to work. Of course it would be optimal for me if your dotnet image was just working...
https://github.com/simonelnahas/FsLexYaccServerTemplate/blob/gpWorkspace/.gitpod.Dockerfile 

```
 dotnet-sdk-2.2 : Depends: aspnetcore-runtime-2.2 (>= 2.2.7) but it is not going to be installed
                  Depends: dotnet-runtime-2.2 (>= 2.2.7) but it is not going to be installed
 dotnet-sdk-3.1 : Depends: dotnet-runtime-3.1 (>= 3.1.1) but it is not going to be installed
                  Depends: aspnetcore-runtime-3.1 (>= 3.1.1) but it is not going to be installed
```

JesterOrNot:
I apologize, the previous commit works though

jankeromnes:
@simonelnahas The problem is that Ubuntu deleted all Ubuntu 19.04 packages, thus breaking `apt` for all environments based on Ubuntu 19.04. That's why we've had to upgrade all Gitpod images to Ubuntu 20.04 LTS last week.

Thus, https://packages.microsoft.com/config/ubuntu/19.04/packages-microsoft-prod.deb will no longer work. That's the reason you're seeing these `Depends: dotnet-runtime-3.1 (>= 3.1.1) but it is not going to be installed` errors.

Furthermore, `dotnet-sdk-2.2` reached end-of-life on 2019-12-23, and doesn't support Ubuntu 20.04, so we've had to remove it from all Gitpod images in https://github.com/gitpod-io/workspace-images/commit/015ce9b3c2dcb847d091dbe153da4252fd836ad9

This leaves us with `dotnet-sdk-3.1`, but:
- https://packages.microsoft.com/config/ubuntu/19.04/packages-microsoft-prod.deb and https://packages.microsoft.com/config/ubuntu/19.10/packages-microsoft-prod.deb do not work for Ubuntu 20.04 (broken dependencies)
- Microsoft has not yet released a https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb

So unfortunately we're a bit stuck (our dotnet image builds will be broken until Microsoft releases a dotnet version for Ubuntu 20.04).

As a workaround, you could base your workspace image on Ubuntu 19.10 (e.g. via `buildpack-deps:eoan`, and then copy/paste all of https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile into your Dockerfile) however please beware that Ubuntu 19.10 will also reach end-of-life pretty soon (in July).

In retrospect it was a mistake to use a non-LTS version of Ubuntu, and starting with Ubuntu 20.04 LTS we'll now always stick to LTS versions.

jankeromnes:
Update: There seems to be a https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb now. 🎉 Microsoft was pretty quick to add it, so ❤️ to them.

Unfortunately, the error `Unable to locate package dotnet-sdk-3.1` is still there.

I've also found "official" instructions for installing .NET Core 3.1 on Ubuntu 20.04, but Google's URL currently gives a 404 page on Microsoft's website. Nevertheless, here is the cached version (thanks Google): https://webcache.googleusercontent.com/search?q=cache:CfIV54IT1KAJ:https://docs.microsoft.com/en-us/dotnet/core/install/linux-package-manager-ubuntu-2004+&cd=1&hl=en&ct=clnk&gl=fr&client=firefox-b-d

They even mention the error we're seeing, and suggest a few workarounds in the section called "Troubleshoot the package manager".

Long story short, .NET Core 3.1 may be coming for Ubuntu 20.04 at any time, which is great news.

@JesterOrNot Could you please rebase this PR on top of `master`? (Sorry for the many conflicting changes I made 😬 it may actually be easier to close this PR and make a new one, copy/pasting your current changes here.)

jankeromnes:
FYI the `gitpod/workspace-dotnet` and `gitpod/workspace-dotnet-vnc` Ubuntu 20.04 images are now fixed: https://github.com/gitpod-io/workspace-images/pull/218

Does `fsharpi` work in them? Or are there more changes required to the dotnet images?

jankeromnes:
Ah, rereading my own comment: https://github.com/gitpod-io/workspace-images/pull/218#issuecomment-621688879 it seems that maybe we need  to install `mono-devel` to fix `fsharpi`? (We already have `mono-complete`, but not `mono-devel` yet.)

PR #202, opened 2020-04-14
Major changes to wsfull

### LINKED ISSUE #136: Don't use Pyenv
Python via Pyenv breaks a lot of integrations, and I would really like it wasn't the default way of doing 3.x.

Proposed solution:
* ~~Keep Pyenv installed~~
* Keep 2.x as is
* ~~Change the *system's version of python3* to 3.8.0~~
* Source build our python

RDIL:
Changes:

* Rely on system python2. Anybody still using those versions should know they are on their own now 👀
* Update python language server - should have better performance
* Remove ruby 2.5 - users can add it themselves if they really need it. this should be the thing helping the size.
* Build python 3 from source !!!!!
* Fixes #136 
* Implements #146
* Implements #194 
* Update rust to 1.42.0

NOTE: one of the changes made is that now, full optimizations are being made to the python install. This makes the build super slow but gives an up to 25% performance boost for python


RDIL:
@jankeromnes @JesterOrNot fyi I moved everything to this PR to stop dazzle from ruining my dreams of completing this migration so please review when you have time :)

RDIL:
@Kreyren @jankeromnes can I have some assistance with apt-get please

full/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM buildpack-deps:disco
+FROM buildpack-deps:bionic
```
RDIL, line 1: Yeah, I am downgrading ubuntu because disco is EOL

jankeromnes:
Hey, many thanks for the hard work @RDIL!

I agree with many of your changes, but am uncertain about others. Maybe it would be more productive to separate changes, thus avoiding a "massive refactor of the entire world (that can never get merged)" effect. 😅

E.g. I'm +1 on:
- the Rust upgrade
- the smaller improvements (e.g. simplifying install or `rm` lines)

The Python stuff probably needs more careful proceeding, and maybe even more discussion, e.g. maybe now that Python 2 is deprecated, we could consider:
- making Python 3 the default again
- maybe even remove Python 2 one day?
- maybe even remove `pyenv` if we do remove Python 2?
- etc

full/Dockerfile
jankeromnes: @RDIL No need, see https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile#L1

RDIL:
Hey @jankeromnes,
The thought for Python is that we have Python 3 as the actual system default, because pyenv breaks a load of stuff, and the system will already have py2, and pyenv is still being installed.

RDIL:
I'll open different PRs though.

full/Dockerfile
```diff
@@ -203,49 +201,65 @@ ENV PATH=$PATH:/home/gitpod/.nvm/versions/node/v${NODE_VERSION}/bin
...
 USER gitpod
 RUN curl -sSL https://rvm.io/mpapis.asc | gpg --import - \
-    && curl -sSL https://rvm.io/pkuczynski.asc | gpg --import - \
+    && curl -fsSL https://rvm.io/pkuczynski.asc | gpg --import - \
     && curl -fsSL https://get.rvm.io | bash -s stable \
     && bash -lc " \
         rvm requirements \
-        && rvm install 2.5 \
         && rvm install 2.6 \
         && rvm use 2.6 --default \
```
jankeromnes, line 252: In 2019, `2.5` still seemed like the most used Ruby version: https://www.jetbrains.com/lp/devecosystem-2019/ruby/

I think it's valuable to have two Ruby versions: the one that's used by the most developers, and the latest one that should be used.

Unless we definitely know that Ruby 2.5 projects work without any change with the latest Ruby too. (E.g., we've removed Java 8 from our images, even though it's still used by most developers, because we assume that Java projects will also work well with latest stable Java.)

RDIL: Fair enough

jankeromnes:
Hey, thanks again @RDIL for the great work! Sorry that the PR got too big to review. I hope that your smaller PRs worked out as you wanted, and if not, please feel free to ping repeatedly until someone reviews your work. 👍 (Anything you've spent some time/efforts/thought on should at least get a review.)

Also yes that goes for your work too @Kreyren. 👍 (And so sorry that my reviewing speed is no match for your implementation speed!)

RDIL:
Don't worry @jankeromnes, I ended up making my own wsfull image that is slimmer and a bit better implemented, [better-wsfull](https://github.com/RDIL/better-wsfull), so the work has not been wasted.

Kreyren:
@RDIL please provide TLDR to what was this about i may be able to implement that in my refactor of workspace-images

RDIL:
@Kreyren see the PR description

Kreyren:
That doesn't provide the expected full abstract in such a way to clearly understand what needs to be done and why

RDIL:
Well tell me what you are confused about and I'll explain it

Kreyren:
>  what needs to be done and why

RDIL:
Well python really does need to not use pyenv.  

Right now, Gitpod's python setup is very baseline, and because it isn't set up in a way the python ecosystem requires, it gets quite messy.  https://github.com/gitpod-io/gitpod/issues/1201 is one of these such issues.

Also, however theia discovers python versions is super odd.  I had to literally delete the system python version because it was trying to use it over the python version I was installing by default.  

This way it is done messes way too much up for development to work.

Kreyren:
In my refactor i am currently relying on downstream to manage the python dependencies would you consider that being a solution?

If not why

PR #212, opened 2020-04-21
[full] Upgrade the Python versions installed by Pyenv (2.7.17 → 2.7.18 and 3.7.7 → 3.8.2)

jankeromnes:
Ubuntu 20.04 comes with Python 3.8.

This causes some VS Code extensions in Gitpod to favour the system version over the Pyenv version, which breaks some assumptions we've made (i.e. we pre-install a bunch of stuff into the Pyenv installs, and have only tested Pyenv versions while doing QA on Gitpod's Python language support).

Upgrading to latest Pyenv versions should (hopefully) solve this.


jankeromnes:
Failed with:

```
python-build: definition not found: 2.7.18
```

Looks like `pyenv` hasn't yet added yesterday's release: https://www.python.org/doc/versions/

Rolling back for now.

full/Dockerfile
```diff
@@ -208,14 +208,14 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
...
-    && pyenv install 3.7.7 \
-    && pyenv global 2.7.17 3.7.7 \
+    && pyenv install 3.8.2 \
+    && pyenv global 2.7.17 3.8.2 \
     && python2 -m pip install --upgrade pip \
     && python3 -m pip install --upgrade pip \
     && python3 -m pip install --upgrade \
         setuptools wheel virtualenv pipenv pylint rope flake8 \
         mypy autopep8 pep8 pylama pydocstyle bandit notebook \
-        python-language-server[all]==0.25.0 twine \
```
RDIL, line 218: Why are you dropping this

jankeromnes: @RDIL I don't think it's used anywhere, so no need to pre-install it.

Do you use it somewhere?

AlexTugarev: @RDIL, we're not using the `@theia/python` extension anymore, which was the reason to add it here in the first place.

RDIL: Okay, understandable.

jankeromnes:
Awesome, thanks for the review @RDIL! 🎉 🙏

PR #213, opened 2020-04-22
[full] Upgrade Ruby 2.5 → 2.5.8 and 2.6 → 2.7.1

jankeromnes:
But keep Ruby 2.5, that way we stay with the latest stable Ruby + the most popular Ruby version ([in 2019, according to JetBrains](https://www.jetbrains.com/lp/devecosystem-2019/ruby/)). Also, I'm looking forward to more recent data for 2020.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,29 +2,35 @@
 
 A curated, chronologically ordered list of notable changes in [Gitpod's default workspace images](https://hub.docker.com/u/gitpod).
 
+## 2020-04-29
+
+- Upgrade RVM's Ruby from `2.5` → `2.5.8` and `2.6` → `2.7.1` [gitpod-io/workspace-images#213](https://github.com/gitpod-io/workspace-images/pull/213)
+- Best practice: Don't stay as `USER root` in `gitpod/workspace-full-vnc` [gitpod-io/workspace-images#215](https://github.com/gitpod-io/workspace-images/pull/215)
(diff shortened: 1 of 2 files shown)


jankeromnes:
Next we should test the Gitpod setup of https://github.com/thepracticaldev/dev.to with this new image, to see if it's fixed.

jankeromnes:
Still fails with `/bin/bash: solargraph: command not found` because `rvm install 2.7` seems to install `2.7.0` and not `2.7.1`:

```
gitpod /workspace/dev.to $ rvm list
Warning! PATH is not properly set up, /workspace/.rvm/bin is not at first place.
         Usually this is caused by shell initialization files. Search for PATH=... entries.
         You can also re-add RVM to your profile by running: rvm get stable --auto-dotfiles
         To fix it temporarily in this shell session run: rvm use .rvm
         To ignore this error add rvm_silence_path_mismatch_check_flag=1 to your ~/.rvmrc file.
   ruby-2.5.7 [ x86_64 ]
   ruby-2.7.0 [ x86_64 ]
 * ruby-2.7.1 [ x86_64 ]

# => - current
# =* - current && default
#  * - default
```

Let's fix this by specifying the patch version as well. (Also, the 2.5 version should ideally be `2.5.8`: https://www.ruby-lang.org/en/downloads/releases/)

jankeromnes:
Opening a `*.rb` file in dev.to now fails with a different error:

```
[Error - 4:32:26 PM] Server initialization failed.
  Message: [ArgumentError] Trying to register Bundler::GemfileError for status code 4 but Bundler::GemfileError is already registered
  Code: -32603 
```

PR #216, opened 2020-04-23
[full] add bash auto-complete for cargo

JesterOrNot:
This adds badly needed autocompletion for the cargo command

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -264,6 +264,7 @@ RUN cp /home/gitpod/.profile /home/gitpod/.profile_orig && \
         rust-analysis \
         rust-src \
     && .cargo/bin/rustup completions bash | sudo tee /etc/bash_completion.d/rustup.bash-completion > /dev/null \
+    && .cargo/bin/rustup completions bash cargo | sudo tee /etc/bash_completion.d/rustup.cargo-bash-completion > /dev/null \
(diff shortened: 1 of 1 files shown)


full/Dockerfile
```diff
@@ -263,7 +263,7 @@ RUN cp /home/gitpod/.profile /home/gitpod/.profile_orig && \
         rls \
         rust-analysis \
         rust-src \
-    && .cargo/bin/rustup completions bash | sudo tee /etc/bash_completion.d/rustup.bash-completion > /dev/null \
+    && .cargo/bin/rustup completions bash cargo | sudo tee /etc/bash_completion.d/rustup.bash-completion > /dev/null \
```
jankeromnes, line 266: Oops, my bad, it seems that `rustup completions bash cargo` actually only outputs completions for `cargo`, and that you still need to use `rustup completions bash` separately.

@JesterOrNot Could you please double-check?

In case both are needed, we could do it like this:

```suggestion
    && .cargo/bin/rustup completions bash | sudo tee /etc/bash_completion.d/rustup.bash-completion > /dev/null \
    && .cargo/bin/rustup completions bash cargo | sudo tee /etc/bash_completion.d/rustup.cargo-bash-completion > /dev/null \
```

JesterOrNot:
Thanks, merging! 🚀

PR #218, opened 2020-04-29
[dotnet][dotnet-vnc] Fix image builds

jankeromnes:
As suggested in https://webcache.googleusercontent.com/search?q=cache:CfIV54IT1KAJ:https://docs.microsoft.com/en-us/dotnet/core/install/linux-package-manager-ubuntu-2004+&cd=1&hl=en&ct=clnk&gl=fr&client=firefox-b-d

### DIFF
--- a/dotnet-vnc/Dockerfile
+++ b/dotnet-vnc/Dockerfile
@@ -1,14 +1,9 @@
 FROM gitpod/workspace-full-vnc:latest
 
-USER root
-RUN wget -q https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb && \
-    dpkg -i packages-microsoft-prod.deb && \
-    rm -rf packages-microsoft-prod.deb && \
-    add-apt-repository universe && \
-    apt-get update && \
-    echo resolvconf resolvconf/linkify-resolvconf boolean false | debconf-set-selections && \
-    DEBIAN_FRONTEND=noninteractive ACCEPT_EULA=Y apt-get -y -o APT::Install-Suggests="true" install \
-        dotnet-sdk-3.1 \
-        fsharp \
-        mono-complete && \
-    rm -rf /var/lib/apt/lists/*
+USER gitpod
+
+# Install .NET Core 3.1 SDK binaries on Ubuntu 20.04
+# Source: https://dev.to/carlos487/installing-dotnet-core-in-ubuntu-20-04-6jh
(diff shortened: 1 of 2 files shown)


jankeromnes:
Still fails with:

```
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package dotnet-sdk-3.1
E: Couldn't find any package by glob 'dotnet-sdk-3.1'
E: Couldn't find any package by regex 'dotnet-sdk-3.1'
```

I guess the fall back is to drop the `.deb` file and install binaries instead, as described in: https://dev.to/carlos487/installing-dotnet-core-in-ubuntu-20-04-6jh

jankeromnes:
New approach. Still left to do:

- [ ] Re-install `fsharp` and `mono-complete`
- [ ] Install `mono-devel` to fix `fsharpi` (i.e. absorb https://github.com/gitpod-io/workspace-images/pull/195)

jankeromnes:
Woohoo, it worked! 🎉 

Screenshot 2020-05-01 at 11 58 29

jankeromnes:
Ok, so the new Ubuntu 20.04 `gitpod/workspace-dotnet*` images now build successfully again.

However, I don't know .NET well enough to know whether it's okay to just install:
- .NET Core SDK 3.1
- `mono-complete`
- `fsharp`

We used to (potentially) install a few extra things, e.g.:

- The previous image used `ACCEPT_EULA=Y apt-get -y -f -o APT::Install-Suggests="true" install` instead of just `apt-get install -y`. Will we miss anything important if we don't install suggested packages?

- In https://github.com/gitpod-io/workspace-images/pull/195, `mono-devel` was also installed (in addition to `mono-complete`). Is that needed, e.g. to fix `fsharpi`?

@JesterOrNot @simonelnahas @jhuizingh Would you have any insights on the questions above?

Otherwise, I guess the next step would be to test the new images, and see if they work with "real" .NET projects:
- `gitpod/workspace-dotnet:branch-jx-fix-dotnet`
- `gitpod/workspace-dotnet-vnc:branch-jx-fix-dotnet`

jankeromnes:
Well, I guess "maybe fixed" is better than "definitely broken", so I'll merge this PR, and we can make any adjustments as follow-up PRs.

Thanks again for your patience and help! Let's make .NET development in Gitpod awesome. 🚀

jhuizingh:
Sorry, meant to respond sooner. I think just the sdk is good for what I'm trying to do.

jankeromnes:
No worries (and in fact, sorry for being hasty, even though it's a prolonged weekend...) Many thanks for your reply! 🙏

PR #241, opened 2020-06-02
[full] Also install 'pip' for system Python 3

jankeromnes:
As discussed with @AlexTugarev: Installing `pip` also for the system version of Python 3 (i.e. not only in Pyenv's Python 3, which is expected to be the default version) makes it easier to use the system version in Gitpod without too many strange errors / warnings / alerts in the IDE.


jankeromnes:
Works as intended:

```
$ python3 -m pip --version
pip 20.1.1 from /home/gitpod/.pyenv/versions/3.8.2/lib/python3.8/site-packages/pip (python 3.8)

$ /usr/bin/python3 -m pip --version
pip 20.0.2 from /usr/lib/python3/dist-packages/pip (python 3.8)
```

(System `pip` version is slightly older, but the point was to make a worst-case scenario "less bad" anyway so that's ok.)

PR #264, opened 2020-07-23
optimization in git clone using --depth

Rajpratik71:
optimize the git clone using --depth flag in term of size of clone
and also in term's of time taken to fetch the files and commit history
of whole repository .

More detail can be found at blog
https://www.atlassian.com/git/tutorials/big-repositories

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/flutter/Dockerfile
+++ b/flutter/Dockerfile
@@ -12,7 +12,7 @@ RUN set -ex; \
 RUN set -ex; \
     mkdir ~/development; \
     cd ~/development; \
-    git clone https://github.com/flutter/flutter.git -b stable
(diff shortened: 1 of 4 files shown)


ret394:
Add a snippet from https://www.atlassian.com/git/tutorials/big-repositories

jankeromnes (approved):
Looks good to me, thanks!

Also, the Flutter build error seems like an unrelated / random network flake, but should remain a watch item until the Flutter image was successfully rebuilt in `master` (but we'll only know when the Go breakage from https://github.com/gitpod-io/workspace-images/pull/350#issuecomment-783187156 is fixed).

PR #266, opened 2020-07-30
Add Dockerfile for Meteor projects

fknipp:


jankeromnes (commented):
Hi @fknipp, many thanks for this Pull Request!

I believe that it's already pretty easy for any Meteor project to install `meteor` in a custom Gitpod Dockerfile (e.g. just like in [meteor/todos#275](https://github.com/meteor/todos/pull/275/files)). Given that, could you please explain what benefits you expect to see from having an official (Gitpod-maintained) `gitpod/workspace-meteor` Docker image?

If there is no significant advantage to having an official image vs allowing anyone to install `meteor` via a custom Gitpod Dockerfile, I would be inclined to keep this repository as small & manageable as possible (i.e. just for languages/tools/stacks that are both highly requested and non-trivial to get working).

Thus, we should probably move this Dockerfile into our [Languages & Frameworks](https://www.gitpod.io/docs/languages-and-frameworks/) docs ([source](https://github.com/gitpod-io/website/tree/master/src/docs/languages)) instead.

This way we'll make sure that anyone interested in using Meteor in Gitpod can find the required steps with a quick online search.

@fknipp do you agree, and would you be interested in doing that?

princerachit:
This is obsolete. Closing

PR #294, opened 2020-10-21
Java 11.0.6.fx-zulu is not available

rajanagori-hl60:
I face the java version issue at the time of docker image build so I change the version and it install the latest version i.e. Java 11.0.8.fx-zulu
link: https://cdn.azul.com/zulu/bin/zulu11.41.23-ca-fx-jdk11.0.8-linux_x64.tar.gz

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -171,7 +171,7 @@ LABEL dazzle/test=tests/lang-java.yaml
 USER gitpod
 RUN curl -fsSL "https://get.sdkman.io" | bash \
  && bash -c ". /home/gitpod/.sdkman/bin/sdkman-init.sh \
-             && sdk install java 11.0.6.fx-zulu \
+             && sdk install java 11.0.8.fx-zulu \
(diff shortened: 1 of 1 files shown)


jankeromnes (approved):
Great catch, many thanks for fixing this! 👍

On a side note, I have now [automated](https://github.com/autofix-dev/autofix/commit/ecca913e8d320877f515ce11a43c15511f2ea66d) the upgrading of such pinned SDKMAN Java versions via [Autofix](https://autofix.dev).

So, if another patch version of SDKMAN Java is released in the future, we should receive an automated upgrade PR. (Note: Autofix runs [twice/week](https://github.com/autofix-dev/cronjobs/blob/78f0b4cb9052d8b53332fad568d35752329e54a1/.circleci/config.yml#L45-L46) [on this repository](https://github.com/autofix-dev/cronjobs/blob/78f0b4cb9052d8b53332fad568d35752329e54a1/.circleci/config.yml#L19-L22).)

PR #305, opened 2020-11-19
Do shallow clone of gecko to clone quickly

championshuttler:
cc @sylvestre 

After this:

image

Before: 

image

### DIFF
--- a/gecko/Dockerfile
+++ b/gecko/Dockerfile
@@ -12,7 +12,7 @@ RUN __RR_VERSION__="5.4.0" \
 # https://firefox-source-docs.mozilla.org/setup/linux_build.html#one-line-bootstrapping
 ENV SHELL=/bin/bash
 RUN sudo apt-get update \
- && git clone --progress https://github.com/mozilla/gecko-dev/ /tmp/gecko \
+ && git clone --progress --single-branch --depth=1 https://github.com/mozilla/gecko-dev/ /tmp/gecko \
  && cd /tmp/gecko \
  && ./mach bootstrap --no-interactive --application-choice=browser \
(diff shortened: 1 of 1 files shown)


sylvestre:
excellent, thanks :+1:

sylvestre:
@jankeromnes ping (quick review I think :)

jankeromnes (approved):
> excellent, thanks 👍

Agreed, many thanks!

> @jankeromnes ping (quick review I think :)

There is one minor thing that makes me want to run the CI first, so I'll do that.

While intuitively it makes sense that shallow-cloning a single branch should be much faster & more efficient than deep-cloning all branches, I vaguely remember some anecdote where a shallow clone took _more_ time than deep clone on GitHub, because deep clone was the "normal case" GitHub optimized for, while shallow clone required extra time-consuming processing on GitHub's side (which might get exacerbated by the large size of `mozilla/gecko-dev`).

So, just to be extra safe, I'll trigger the CI once, then merge if results look good.

For comparison, `workspace-gecko` is currently built in 34min 49s (average from the last 10 builds) and clones about 2.87 GiB (last build).

sylvestre:
Suspense :)

championshuttler:
OMG, did we just decrease build  time by 10 minutes? 🦊

jankeromnes:
> For comparison, `workspace-gecko` is currently built in 34min 49s (average from the last 10 builds) and clones about 2.87 GiB (last build).

The CI took 20min 59s to build the `workspace-gecko` image, and only cloned 536.67 MiB.

So yes, your PR appears to shave off ~14min of build time and 2.35 GiB of cloned objects! 🚀 🥇 ✨

Many thanks @championshuttler 🙏

PR #313, opened 2020-12-02
Remove Workaround for `runc` path bug

meysholdt:
fix is here: https://github.com/gitpod-io/gitpod/pull/2379


jankeromnes (approved):
Looks good to me 👍 Thanks!

PR #315, opened 2020-12-03
Remove more stuff. Should save more than 5g

sylvestre:


gecko/Dockerfile
```diff
@@ -35,8 +35,8 @@ ENV PATH $PATH:$HOME/.git-cinnabar
 # Inject the path update to the top as bash completion needs to call rustc
 RUN sed -i '1s|^|PATH=\"\$PATH:$HOME/.cargo/bin/\"\n|' $HOME/.bashrc
 
-# Delete go directories
-RUN rm -rf go-packages go
+# Delete useless directories for Firefox
+RUN rm -rf go-packages go .gem .subversion .rvm
```
jankeromnes, line 39: Sadly, removing stuff here will not have the desired effect.

Docker captures the entire file system state after each `RUN` instruction, and saves it into a dedicated "layer". Modifying or deleting files in a different `RUN` instruction later will just create a new layer that "masks" (or "hides") the previous layers, but the large files will remain present in Docker's layer history (just like if you add a giant file with `git` or `hg`, and then later delete the file in a new commit, the large file will remain in the repo history and continue to cause long cloning times).

In order to fully remove files from Docker images (and make the images actually lighter), we need to find in which layer or `RUN` instruction the large files are being created, and change that `RUN` instruction.

In other words:

```Dockerfile
RUN wget 30GB.tar.gz
RUN rm 30GB.tar.gz
```

will still produce a 30GB+ image.

sylvestre: Oh, yeah, you see how much I am an expert on this ;)

What is your opinion about multistage builds ?
https://docs.docker.com/develop/develop-images/multistage-build/

jankeromnes:
Meilleurs voeux pour 2021 @sylvestre! 🎆 🥳 

As discussed above, this isn't the correct approach to save space on Dockerfiles. However, https://github.com/gitpod-io/workspace-images/pull/318 already removed a lot of unnecessary stuff, so we can call this done. 🙌

sylvestre:
Bof: go and homebrew are still pretty big (more than one g) and useless for Firefox dev :)

jankeromnes:
Indeed. In that case, it would be good to have a way to disable specific Dazzle layers that you don't need (feature not implemented yet).

However, Homebrew is required to have the latest version of CMake. Does Firefox use CMake?

Also, today I learned that there are two `.go` files in the Firefox source tree: 😋
- https://searchfox.org/mozilla-central/source/toolkit/crashreporter/google-breakpad/src/tools/mac/upload_system_symbols/upload_system_symbols.go
- https://searchfox.org/mozilla-central/source/toolkit/crashreporter/google-breakpad/src/tools/mac/upload_system_symbols/arch_reader.go

PR #318, opened 2020-12-08
Make gitpod/workspace-full image lighter by removing unnecessary/duplicate stuff

jankeromnes:
Fixes https://github.com/gitpod-io/workspace-images/issues/124

This PR currently reduces `gitpod/workspace-full` from `7.33 GB` to `6.46 GB`.

PR #319, opened 2020-12-10
Upgrade dotnet stacks to .NET 5.0

sr229:
This upgrades the .NET stacks to .NET 5.0. This also gets rid of the awful  F# Mono installation we used to do as F# is now part of the .NET 5.0 stack.

### DIFF
--- a/dotnet-vnc/Dockerfile
+++ b/dotnet-vnc/Dockerfile
@@ -2,17 +2,8 @@ FROM gitpod/workspace-full-vnc:latest
 
 USER gitpod
 
-# Install .NET Core 3.1 SDK binaries on Ubuntu 20.04
+# Install .NET Core 5.0 SDK binaries on Ubuntu 20.04
 # Source: https://dev.to/carlos487/installing-dotnet-core-in-ubuntu-20-04-6jh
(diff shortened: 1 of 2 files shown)


jankeromnes:
Wow, there is already a .NET Core 5.0! 👀 Many thanks for making the upgrade @sr229! 🙌

Normally we would tend to stick with the latest available LTS, unless there is a significant advantage (and/or negligible risk) to upgrading. In this case, an improved F# setup does sound like a notable advantage.

Out of curiosity, what is awful about the furrent F# Mono installation? (Apart from the fact that it [randomly breaks](https://github.com/gitpod-io/workspace-images/pull/318#issuecomment-742408545).)

jankeromnes:
Gah, sorry, while trying to push your commit to a branch, I accidentally pushed to `master`. 🤦

And now I can't re-open this PR.

sr229:
> Wow, there is already a .NET Core 5.0! 👀 Many thanks for making the upgrade @sr229! 🙌
> 
> Normally we would tend to stick with the latest available LTS, unless there is a significant advantage (and/or negligible risk) to upgrading. In this case, an improved F# setup does sound like a notable advantage.
> 
> Out of curiosity, what is awful about the furrent F# Mono installation? (Apart from the fact that it [randomly breaks](https://github.com/gitpod-io/workspace-images/pull/318#issuecomment-742408545).)

Mono is actually quite hated by some people (due to its Framework roots) and this is the main reason why .NET 5.0 was pushed since it basically deprecated both Mono, .NET Core and .NET Framework, from now on there's only one .NET implementation to maintain for them.

PR #322, opened 2020-12-17
Add golangci-lint to workspace-full

ArthurSens:
As a preparation for our adoption of `golanci-lint` as our linter.


ArthurSens:
@csweichel Do we want the linter in other images too?

csweichel:
I don't think we need to add this to workspace full. Adding it to the image we use in gitpod-io/gitpod would be enough.

ArthurSens:
Ops my bad, without looking I just assumed we used the workspace-full :man_facepalming: 

Do you know where exactly do I change the image: `eu.gcr.io/gitpod-core-dev/dev/dev-environment`?

ArthurSens:
Closing since this is not needed

PR #331, opened 2021-01-14
[full] Invalidate Dazzle layer cache to force-update to latest stable Docker

jankeromnes:
Fixes: https://community.gitpod.io/t/docker-20-10-support-on-saas/2499

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -263,6 +263,8 @@ RUN bash -lc "cargo install cargo-watch cargo-edit cargo-tree"
 LABEL dazzle/layer=tool-docker
 LABEL dazzle/test=tests/tool-docker.yaml
 USER root
+# Dazzle does not rebuild a layer until one of its lines are changed. Increase this counter to rebuild this layer.
+ENV TRIGGER_DOCKER_REBUILD=1
 # https://docs.docker.com/engine/install/ubuntu/
 RUN curl -o /var/lib/apt/dazzle-marks/docker.gpg -fsSL https://download.docker.com/linux/ubuntu/gpg \
     && apt-key add /var/lib/apt/dazzle-marks/docker.gpg \


tnir (approved):

jankeromnes:
Woops, I seem to have forgotten about this. Will re-trigger the CI, then test the resulting image with some repository that needs Docker.

jankeromnes:
Ah, seems like testing was a good call, because the default WordPress / docker-compose example seems to break after this update.

**Before this PR** (https://gitlab.com/jankeromnes/wordpress):

```
$ docker version
Client: Docker Engine - Community
 Version:           19.03.14
 API version:       1.40
 Go version:        go1.13.15
 Git commit:        5eb3275d40
 Built:             Tue Dec  1 19:20:26 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.14
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       5eb3275d40
  Built:            Tue Dec  1 19:18:53 2020
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.3.9
  GitCommit:        ea765aba0d05254012b0b9e595e995c09186427f
 gitpod:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

```
$ docker-compose up
Starting wordpress_db_1 ... done
Starting wordpress_wordpress_1 ... done
Attaching to wordpress_db_1, wordpress_wordpress_1
# ...
```

✅ It works!

**After this PR** (https://gitlab.com/jankeromnes/wordpress/-/tree/test-workspace-images/pull/331):

```
$ docker version
Client: Docker Engine - Community
 Version:           20.10.2
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        2291f61
 Built:             Mon Dec 28 16:17:43 2020
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.2
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       8891c58
  Built:            Mon Dec 28 16:15:19 2020
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 gitpod:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

```
$ docker-compose up
Starting wordpress_db_1 ... error

ERROR: for wordpress_db_1  Cannot start service db: OCI runtime create failed: container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: write sysctl key net.ipv4.ip_unprivileged_port_start: open /proc/sys/net/ipv4/ip_unprivileged_port_start: read-only file system: unknown

ERROR: for db  Cannot start service db: OCI runtime create failed: container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: write sysctl key net.ipv4.ip_unprivileged_port_start: open /proc/sys/net/ipv4/ip_unprivileged_port_start: read-only file system: unknown
ERROR: Encountered errors while bringing up the project.
```

❌ Fail

csweichel:
[Moby introduced a change](https://github.com/moby/moby/blob/470ae8422fc6f1845288eb7572253b08f1e6edf8/daemon/oci_linux.go#L774-L777) that assumes one can just write those sysctls. We mask `/proc/sys` in general, because that's what runc/kubernetes do by default. 

We'd either have to raise a PR towards Moby so that 
- it doesn't fail if those sysctls fail (they're primarily for convenience)
- it checks if we have `CAP_NET_BIND_SERVICE` prior to writing the sysctl
or we don't mask `/proc/sys/net/ipv4/ip_unprivileged_port_start` which would make the proc masking and future AppArmor rules considerably more complicated.

lioramilbaum:
Have you considered dropping docker in favor of other technologies which provides similar containerization capabilities?

jankeromnes:
Good point @lioramilbaum! [gitpod.io](https://gitpod.io) actually uses `containerd` behind the scenes, but we ship `docker` inside user workspaces because that's a popular demand (e.g. to conveniently `docker build` or `docker run` in your workspace).

jankeromnes:
Superseded by https://github.com/gitpod-io/workspace-images/pull/362

PR #349, opened 2021-02-18
Use CUSTOM_XVFB_WxHxD for initial screen resolution

markfirmware:


jankeromnes:
I've just triggered a CircleCI build for this PR, so that we can test it.

Once through, the following image should be tested:

```
gitpod/workspace-full-vnc:branch-pr-349
```

jankeromnes:
I've created https://github.com/jankeromnes/gitpod-custom-novnc-resolution to test this.

- [ ] First without customization
- [ ] Then with a `CUSTOM_XVFB_WxHxD` env var in https://gitpod.io/settings/ (filtered just for that repo)

jankeromnes:
Ah, maybe `1280x720x16` isn't a supported value?

Trying with `1280x1024x8` works (that's one of the example values in https://www.x.org/releases/X11R7.6/doc/man/man1/Xvfb.1.xhtml):

Screenshot 2021-02-24 at 11 03 06

Screenshot 2021-02-24 at 11 01 53

jankeromnes:
`1280x1024x32` also doesn't work:

Screenshot 2021-02-24 at 11 26 14

It would be great to:
- [ ] Know which value combinations are "acceptable"
- [ ] Somehow get the error logs of https://github.com/gitpod-io/workspace-images/pull/349/files#diff-d14b1a495489c1fb74081410e09dfc13bb2390d76b4b8030fc3d1f004917bf1cR7 when it fails

jankeromnes (approved):
But then again, these only become concerns when you actually set a custom value for `CUSTOM_XVFB_WxHxD`, at which point you either already know what you're doing, or you can at least find a solution relatively easily.

So, I'm happy to merge this Pull Request as is, and leave the aforementioned improvements as potential follow-ups.

PR #352, opened 2021-02-22
[Full] fix golang build

gengjiawen:
## why change user to root
Permission issues.

## why some packages are deleted
They are not maintained anymore, and won't build in go16

@jankeromnes

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -134,7 +134,7 @@ RUN curl -fsSL https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.t
 # install VS Code Go tools from https://github.com/Microsoft/vscode-go/blob/0faec7e5a8a69d71093f08e035d33beb3ded8626/src/goInstallTools.ts#L19-L45
     go get -u -v \
         github.com/mdempsky/gocode \
-        github.com/uudashr/gopkgs/cmd/gopkgs \
+        github.com/uudashr/gopkgs/cmd/gopkgs@v2 \
         github.com/ramya-rao-a/go-outline \
         github.com/acroca/go-symbols \
         golang.org/x/tools/cmd/guru \
@@ -150,7 +150,6 @@ RUN curl -fsSL https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.t
         winterdrache.de/goformat/goformat \
         golang.org/x/lint/golint \
         github.com/cweill/gotests/... \
-        github.com/alecthomas/gometalinter \
         honnef.co/go/tools/... \
         github.com/mgechev/revive \
         github.com/sourcegraph/go-langserver \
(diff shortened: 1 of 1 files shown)


jankeromnes:
Many thanks for working on fixing our build @gengjiawen! 💯 

> ## why change user to root
> Permission issues.

Hmm, I'm not convinced this is the right fix.

In these images, we're setting things up for user `gitpod`. Anything we create as `root` has the risk of not being accessible to use `gitpod`. And it has happened many times in the past that users switched to `USER root`, and then accidentally created some files/directories that they could no longer access as user `gitpod` later on when the workspaces started.

That's why typically, we try to always use `USER gitpod`, and only for these specific commands that do need elevated privileges, we prefix just these with `sudo`.

> ## why some packages are deleted
> They are not maintained anymore, but won't build in go16

Aha, thanks! This makes sense to me. But I'll also ping someone who knows more about Go, like @csweichel, to get a second opinion.

Also, I'll now trigger a CircleCI build for this Pull Request.

gengjiawen:
Looks like go build passed. Failed on other steps in CI flow.

jankeromnes:
> Looks like go build passed.

Indeed! Many thanks for the quick fix. 🙏

> Failed on other steps in CI flow.

True, it seems like the Dazzle test expecting no root-owned filed in HOME failed:

```
   step 00      testing "it should not have root-owned files in the home directory"command=[find /home/gitpod -not "(" -user gitpod -and -group gitpod ")" -print -quit] 
                - status == 0                               
                - stdout.trim().length == 0                 
                failed                                      message="assertion failed: stdout.trim().length == 0" result="&test.RunResult{Stdout: []byte("/home/gitpod/go\n"), Stderr: []byte("bash: cannot set terminal process group (-1): Inappropriate ioctl for device\nbash: no job control in this shell\n"), }" 
```

Maybe let's try without the `USER root` change?

gengjiawen:
> Maybe let's try without the USER root change? EDIT: I've just pushed 5cf868a to your branch & re-triggered the CI.

Looks like CI passed.

jankeromnes (approved):
From my side, these changes look good, thanks! (I just have no idea which Go dependencies are useful or not.)

jankeromnes:
Second opinion from Slack:

> the less built-in Go packages we make others depend on the better.
> Those tools are in there to make the Go extension work nicer OOTB, but AFAIK that extension can take care of itself :slightly_smiling_face:
> Didn’t test, but at first glance :white_check_mark:  LGTM

That's good enough for me, as extensions can install their own packages, and if a common extension no longer works out-of-the-box we can easily fix that again in a follow-up change.

jankeromnes:
Note: I've cancelled the CircleCI master build, because we don't want to deploy new versions of `workspace-full` too frequently, and I'd like to include a few more contributions into this release if possible:
- https://github.com/gitpod-io/workspace-images/pull/349
- https://github.com/gitpod-io/workspace-images/pull/353
- https://github.com/gitpod-io/workspace-images/pull/263

jimmybrancaccio:
Sorry for the double comments, I just wanted to make this visible:

For what it's worth I literally made these changes myself yesterday and was going to submit an issue to have them updated here as well, but it looks like you beat me to the punch. 😝  That said, it seems the replacement for github.com/alecthomas/gometalinter is github.com/golangci/golangci-lint so that's what I changed it to on my end. Not sure if y'all want to do the same.

jankeromnes:
FYI, there is now a relevant-looking warning in Gitpod:

Screenshot 2021-02-26 at 14 58 29

gengjiawen:
> FYI, there is now a relevant-looking warning in Gitpod:
> 
> Screenshot 2021-02-26 at 14 58 29
> 
> EDIT: When you click on "Show", this gets printed in the "Go" extension "Output" console:
> 
> ```
> Below tools are needed for the basic features of the Go extension.
> gocode-gomod
> ```

I am using vscode as default IDE, so far it looks good. Maybe only an issue for Theia.

image

PR #362, opened 2021-03-04
Fix issues with apache2 and update docker-compose

### LINKED ISSUE #339: Update composer (php package manager) to v2
Upgrade the Docker Image to use the latest composer version, this is currently v2.0.9

https://github.com/composer/composer

aledbf:
fixes https://github.com/apolopena/gitpod-laravel8-starter/issues/46
fixes https://github.com/gitpod-io/gitpod/issues/3352
fixes #339

Update docker compose to latest stable version. Current version [1.25.0](https://github.com/docker/compose/releases/tag/1.25.0) was released in Nov 2019.

Screenshot from 2021-03-04 15-19-49

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -3,6 +3,8 @@ FROM buildpack-deps:focal
 COPY install-packages /usr/bin
 
 ### base ###
+ARG DEBIAN_FRONTEND=noninteractive
+
 RUN yes | unminimize \
     && install-packages \
         zip \
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -1,4 +1,4 @@
-# Dazzle do not supports ARG TRIGGER_REBUILD=1
+# Dazzle do not supports ARG TRIGGER_REBUILD=2
 FROM gitpod/workspace-base:latest
 
 RUN echo "ws full starts"
@@ -209,7 +209,7 @@ RUN bash -lc "cargo install cargo-watch cargo-edit cargo-tree"
 LABEL dazzle/layer=tool-docker
 LABEL dazzle/test=tests/tool-docker.yaml
 USER root
-ENV TRIGGER_REBUILD=1
+ENV TRIGGER_REBUILD=2
 # https://docs.docker.com/engine/install/ubuntu/
 RUN curl -o /var/lib/apt/dazzle-marks/docker.gpg -fsSL https://download.docker.com/linux/ubuntu/gpg \
     && apt-key add /var/lib/apt/dazzle-marks/docker.gpg \
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -1,3 +1,4 @@
+# Dazzle do not supports ARG TRIGGER_REBUILD=1
 FROM gitpod/workspace-base:latest
 
 RUN echo "ws full starts"
@@ -23,7 +24,7 @@ RUN curl -o /var/lib/apt/dazzle-marks/llvm.gpg -fsSL https://apt.llvm.org/llvm-s
 LABEL dazzle/layer=tool-nginx
 LABEL dazzle/test=tests/lang-php.yaml
 USER root
-ENV TRIGGER_REBUILD=1
+ENV TRIGGER_REBUILD=2
 RUN install-packages \
         apache2 \
         nginx \
@@ -208,6 +209,7 @@ RUN bash -lc "cargo install cargo-watch cargo-edit cargo-tree"
 LABEL dazzle/layer=tool-docker
 LABEL dazzle/test=tests/tool-docker.yaml
 USER root
+ENV TRIGGER_REBUILD=1
 # https://docs.docker.com/engine/install/ubuntu/
 RUN curl -o /var/lib/apt/dazzle-marks/docker.gpg -fsSL https://download.docker.com/linux/ubuntu/gpg \
     && apt-key add /var/lib/apt/dazzle-marks/docker.gpg \
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -46,7 +46,7 @@ RUN install-packages \
         php-tokenizer \
         php-xml \
         php-zip \
-    && mkdir -p /var/run/nginx /var/log/apache2 /var/log/nginx/ \
+    && mkdir -p /var/run/nginx \
     && ln -s /etc/apache2/mods-available/rewrite.load /etc/apache2/mods-enabled/rewrite.load \
     && chown -R gitpod:gitpod /etc/apache2 /var/run/apache2 /var/lock/apache2 /var/log/apache2 \
     && chown -R gitpod:gitpod /etc/nginx /var/run/nginx /var/lib/nginx/ /var/log/nginx/
@@ -212,7 +212,11 @@ USER root
 RUN curl -o /var/lib/apt/dazzle-marks/docker.gpg -fsSL https://download.docker.com/linux/ubuntu/gpg \
     && apt-key add /var/lib/apt/dazzle-marks/docker.gpg \
     && add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" \
-    && install-packages docker-ce=5:19.03.15~3-0~ubuntu-focal docker-ce-cli=5:19.03.15~3-0~ubuntu-focal containerd.io slirp4netns docker-compose
+    && install-packages docker-ce=5:19.03.15~3-0~ubuntu-focal docker-ce-cli=5:19.03.15~3-0~ubuntu-focal containerd.io slirp4netns
+
+RUN curl -o /usr/local/bin/docker-compose -fsSL https://github.com/docker/compose/releases/download/1.28.5/docker-compose-Linux-x86_64 \
+    && chmod +x /usr/local/bin/docker-compose
+
 # https://github.com/wagoodman/dive
 RUN curl -o /tmp/dive.deb -fsSL https://github.com/wagoodman/dive/releases/download/v0.9.2/dive_0.9.2_linux_amd64.deb \
     && apt install /tmp/dive.deb \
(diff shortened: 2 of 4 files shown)


apolopena (commented):
Ah so the fix is to let `apache2` and `ngix` create their own log directories implicity?

aledbf:
> Ah so the fix is to let `apache2` and `ngix` create their own log directories implicity?

Yes, the mkdir command was added before the final cleanup of the `install-packages` script that removed `/var/log`.
This is not an issue now

apolopena:
> > Ah so the fix is to let `apache2` and `ngix` create their own log directories implicity?
> 
> Yes, the mkdir command was added before the final cleanup of the `install-packages` script that removed `/var/log`.
> This is not an issue now

Thanks for clarificatiom. I always learn better when I understand the nature of the fixes.

aledbf:
@jankeromnes can we release a new version of the images?

aledbf:
The CI failure is due to dazzle using the workspace-base image from the docker registry. 
@jankeromnes can we push workspace-base to fix the problem?

jankeromnes:
Hi @aledbf!

New versions of all images are automatically published to Docker Hub by CircleCI every time we merge something to `master`.

So `workspace-base` has already been published a few times: https://hub.docker.com/r/gitpod/workspace-base/tags

jankeromnes:
I can look into the CI failure more deeply next week if that helps.

aledbf:
@jankeromnes for some reason dazzle is using a base image from eleven days ago
`building image                              dockerfile="dazzle___base.Dockerfile" tag="gitpod/dazzle-wsfull-build:base-cc44bcce9025d17eabc338020bad8a347af716c322f11d11b72193be8c6ef210" 
              `

base/Dockerfile
```diff
@@ -3,6 +3,8 @@ FROM buildpack-deps:focal
 COPY install-packages /usr/bin
 
 ### base ###
+ARG DEBIAN_FRONTEND=noninteractive
```
jankeromnes, line 6: ⚠️ `ENV` variables carry over into Gitpod workspaces. Maybe it would be surprising to users if the Debian frontend is permanently non-interactive (even in their interactive terminals)?

I typically set this inline for each `apt-get install` command to work around this. Alternatively, we could unset it again at the end of `base/Dockerfile` or `full/Dockerfile`.

EDIT: Is this actually even needed here? `install-packages` already has it locally.

full/Dockerfile
```diff
@@ -1,3 +1,4 @@
+# Dazzle do not supports ARG TRIGGER_REBUILD=2
```
jankeromnes, line 1: Dazzle supports `ENV`, but decides to fail early when it sees `ARG`.

I've already fixed that in https://github.com/gitpod-io/dazzle/pull/6

full/Dockerfile
```diff
@@ -208,11 +209,16 @@ RUN bash -lc "cargo install cargo-watch cargo-edit cargo-tree"
 LABEL dazzle/layer=tool-docker
 LABEL dazzle/test=tests/tool-docker.yaml
 USER root
+ENV TRIGGER_REBUILD=2
```
jankeromnes, line 212: I don't think you need to use the `TRIGGER_REBUILD` trick when you also modify any line between `LABEL dazzle/layer=tool-docker` and the next `LABEL dazzle/layer` (as you did below).

jankeromnes (commented):
Many thanks for this PR!

Note: You can use `ENV` instead of `ARG` to make Dazzle happy.

jankeromnes:
> @jankeromnes for some reason dazzle is using a base image from eleven days ago
> `building image dockerfile="dazzle___base.Dockerfile" tag="gitpod/dazzle-wsfull-build:base-cc44bcce9025d17eabc338020bad8a347af716c322f11d11b72193be8c6ef210"`

Hmm, I'm not sure why that is.

- The `workspace-base` CircleCI job should build the `gitpod/workspace-base` image, then export it to `workspace-base.tar`:

https://github.com/gitpod-io/workspace-images/blob/81a67e22ec616fcf2f102541e8ceb6dcf929c213/.circleci/config.yml#L13-L15

- Then, the `workspace-full` job loads that tarfile again, and should build the `gitpod/workspace-full` image based on that:

https://github.com/gitpod-io/workspace-images/blob/81a67e22ec616fcf2f102541e8ceb6dcf929c213/.circleci/config.yml#L31-L33

Maybe there is some issue in the image tags and/or the import/export mechanism that causes Dazzle to pull a different image than what's restored from the previous job?

jankeromnes:
### workspace-base

`./.circleci/build_image.sh base/Dockerfile gitpod/workspace-base` [translates to](https://github.com/gitpod-io/workspace-images/blob/81a67e22ec616fcf2f102541e8ceb6dcf929c213/.circleci/build_image.sh#L9-L13):

```
docker build -t gitpod/workspace-base -f base/Dockerfile base/
```

I expect that to produce a `gitpod/workspace-base:latest` image (because `:latest` is the default tag).

Then, when building a non-`master` branch or PR, we [re-tag & push the image](https://github.com/gitpod-io/workspace-images/blob/81a67e22ec616fcf2f102541e8ceb6dcf929c213/.circleci/build_image.sh#L22-L26) (but I expect that to leave the `:latest` tag intact, creating a second `:branch-XYZ` tag instead):

```
docker tag gitpod/workspace-base gitpod/workspace-base:branch-XYZ
docker push gitpod/workspace-base:branch-XYZ
```

(When it's the `master` branch, it [re-tags (no-op?) and pushes `:latest`](https://github.com/gitpod-io/workspace-images/blob/81a67e22ec616fcf2f102541e8ceb6dcf929c213/.circleci/build_image.sh#L28-L30), and also [tags & pushes `:commit-XYZ`](https://github.com/gitpod-io/workspace-images/blob/81a67e22ec616fcf2f102541e8ceb6dcf929c213/.circleci/build_image.sh#L31-L34).)

Finally, it exports all tags (either `:latest` and `:branch-XYZ`, or `:latest` and `:commit-XYZ`) as [workspace-base.tar](https://github.com/gitpod-io/workspace-images/blob/81a67e22ec616fcf2f102541e8ceb6dcf929c213/.circleci/config.yml#L15):

```
docker save gitpod/workspace-base -o workspace-base.tar
```

### workspace-full

We [load the tarred image](https://github.com/gitpod-io/workspace-images/blob/81a67e22ec616fcf2f102541e8ceb6dcf929c213/.circleci/config.yml#L32) from the previous job:

```
docker load -i workspace-base.tar
```

I expect this to restore all tags: either `:latest` and `:branch-XYZ`, or `:latest` and `:commit-XYZ`, but `gitpod/workspace-base:latest` should absolutely be restored locally at this point.

Then `./.circleci/dazzle_build_image.sh full/Dockerfile gitpod/workspace-full workspace-full.tar` [translates to](https://github.com/gitpod-io/workspace-images/blob/81a67e22ec616fcf2f102541e8ceb6dcf929c213/.circleci/dazzle_build_image.sh#L14-L29):

```
cd full/
dazzle build --repository gitpod/dazzle-wsfull-build --output-test-xml results.xml -t gitpod/workspace-full:build-branch-XYZ -f Dockerfile .
```

I'm not exactly sure how Dazzle works at this point, but since the [first line of `full/Dockerfile`](https://github.com/gitpod-io/workspace-images/blob/81a67e22ec616fcf2f102541e8ceb6dcf929c213/full/Dockerfile#L1) is `FROM gitpod/workspace-base:latest`, I would expect Dazzle to just use the local image & not try to pull something else from Docker Hub.

base/Dockerfile
aledbf: ARG means this is valid only during docker build

full/Dockerfile
aledbf: This is the way I found to avoid using an old workspace-base cache during workspace-full build

aledbf:
@jankeromnes I can build all the images locally without any issue using `docker build`

jankeromnes:
@csweichel Given a `gitpod/workspace-base:latest` image available locally, and a Dockerfile using `FROM gitpod/workspace-base:latest`, will this Dazzle command use the local image or pull something different?

```
dazzle build --repository gitpod/dazzle-wsfull-build --output-test-xml results.xml -t gitpod/workspace-full:build-branch-XYZ -f Dockerfile .
```

aledbf:
>  Dazzle command use the local image or pull something different?

It pulls from the docker registry.

aledbf:
@jankeromnes after today's change in master, the PR is passing.

jankeromnes:
@aledbf Thanks! I guess we still have the problem that, in a PR, `workspace-full` will not actually build upon any `workspace-base` changes you make, but on the last `workspace-base` that was published to Docker Hub.

But that's out-of-scope for this PR. Will take a look (thanks a lot!)

base/Dockerfile
jankeromnes: Today I learned:

> Unlike an `ARG` instruction, `ENV` values are always persisted in the built image.

jankeromnes (approved):
Looks good to me, thanks! Please feel free to merge this.

- If we want the changes to go live as soon as possible, you can let CircleCI rebuild & republish all images after the merge (we did another release yesterday, but it's probably okay to redo one today -- we just shouldn't do that too frequently, ideally one release every 2 weeks seems good)

- If it's okay to delay the changes for a bit, please merge, then log into CircleCI and cancel the triggered `master` pipeline. That way, we can collect a few more changes on `master` over the next week or so, then eventually let a pipeline complete (e.g. in one or two weeks) to deploy all changes together.

apolopena:
Good job guys!
I did some live tests to be able to close https://github.com/apolopena/gitpod-laravel8-starter/issues/46
I can confirm that the apache2 logs are created and apache2 now can be started without any workarounds. In fact any workarounds that anyone has implement in their projects that create the apache logs from `.gitpod.Dockerfile` will break the workspace image build so they should be removed in favor of this fix.

Issue #363, opened 2021-03-04
Update npm


FeixuRuins:
![image](https://user-images.githubusercontent.com/66232834/110005228-3fd24800-7ce6-11eb-9874-5caeb32d8034.png)

akosyakov:
It looks that npm is coming together with Node.js and configured by nvm here: https://github.com/gitpod-io/workspace-images/blob/0d86f18fcb06832838d2cb77636d2c1a6a7dbe6c/full/Dockerfile#L137

gengjiawen:
The image already using latest npm, you can see 
https://github.com/gitpod-io/workspace-images/blob/0d86f18fcb06832838d2cb77636d2c1a6a7dbe6c/full/Dockerfile#L142

`workspace-full` has npm 7. You are probably using custom docker image.

![image](https://user-images.githubusercontent.com/3759816/110264447-49182900-7ff4-11eb-9346-69e0df680d7f.png)

FeixuRuins:
My workspace must be outdated then. I'm using https://github.com/gitpod-io/sveltejs-template, created a few weeks ago.

smacintyre:
It causes a lot lot of problems if you mix the wrong NPM version with Node. The Node LTS should be using NPM 6.x, not 7. I've filed an issue #368 to fix this. NPM 7 should not be installed.

PR #364, opened 2021-03-06
Change NODE_VERSION to 15.11.0

FeixuRuins:
Comes with npm 7.6.0 - #363

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -134,7 +134,7 @@ ENV GRADLE_USER_HOME=/workspace/.gradle/
 LABEL dazzle/layer=lang-node
 LABEL dazzle/test=tests/lang-node.yaml
 USER gitpod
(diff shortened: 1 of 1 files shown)


akosyakov:
15.11.0 is not LTS, proper one is 14.16.0

If someone wants to use unstable Node.js or newer npm then it should be configured in the project image: https://www.gitpod.io/docs/languages/javascript/#node-versions

gengjiawen:
> 15.11.0 is not LTS, proper one is 14.16.0
> 
> If someone wants to use unstable Node.js or newer npm then it should be configured in the project image: [gitpod.io/docs/languages/javascript/#node-versions](https://www.gitpod.io/docs/languages/javascript/#node-versions)

Another reason: https://github.com/gitpod-io/workspace-images/issues/363#issuecomment-792407702

Issue #368, opened 2021-03-12
Wrong version of NPM installed; breaking prebuilds


smacintyre:
There is something wrong with your build process. NPM 7 keeps getting installed when it shouldn't be. This is a repeat of #358. The build tracks the Node LTS which always uses NPM 6.x. NPM 7 represent major changes. Because the installed Node has been update to 14.16.0 the previous fix in #358 we must first add a different version, the uninstall 14.16.0, and then reinstall 14.16.0. 

As long as the image is tracking the LTS, it should be using the bundled NPM -- which is version 6.

smacintyre:
From a fresh launch / work around to fix the bad NPM install:

```sh
$ nvm ls
->     v14.16.0
default -> 14.16.0 (-> v14.16.0)
node -> stable (-> v14.16.0) (default)
stable -> 14.16 (-> v14.16.0) (default)
iojs -> N/A (default)
unstable -> N/A (default)
lts/* -> lts/fermium (-> v14.16.0)
lts/argon -> v4.9.1 (-> N/A)
lts/boron -> v6.17.1 (-> N/A)
lts/carbon -> v8.17.0 (-> N/A)
lts/dubnium -> v10.24.0 (-> N/A)
lts/erbium -> v12.21.0 (-> N/A)
lts/fermium -> v14.16.0

$ npm -v
7.5.6

$ nvm install --lts
Installing latest LTS version.
v14.16.0 is already installed.
Now using node v14.16.0 (npm v7.5.6)

$ nvm install v14.15.0
Downloading and installing node v14.15.0...
Downloading https://nodejs.org/dist/v14.15.0/node-v14.15.0-linux-x64.tar.xz...
######################################################################################################################################################################################################## 100.0%
Computing checksum with sha256sum
Checksums matched!
Now using node v14.15.0 (npm v6.14.8)

$ nvm uninstall v14.16.0
Uninstalled node v14.16.0

$ nvm install --lts
Installing latest LTS version.
Downloading and installing node v14.16.0...
Local cache found: ${NVM_DIR}/.cache/bin/node-v14.16.0-linux-x64/node-v14.16.0-linux-x64.tar.xz
Checksums match! Using existing downloaded archive ${NVM_DIR}/.cache/bin/node-v14.16.0-linux-x64/node-v14.16.0-linux-x64.tar.xz
Now using node v14.16.0 (npm v6.14.11)

$ npm -v
6.14.11

$ nvm uninstall v14.15.0
[...]

PR #372, opened 2021-03-17
[full] Upgrade NVM to v0.37.2

jankeromnes:
Side-effect: nukes Node.js Dazzle layer cache to fix https://github.com/gitpod-io/workspace-images/issues/368

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -136,7 +136,7 @@ LABEL dazzle/layer=lang-node
 LABEL dazzle/test=tests/lang-node.yaml
 USER gitpod
 ENV NODE_VERSION=14.16.0
-RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.37.0/install.sh | PROFILE=/dev/null bash \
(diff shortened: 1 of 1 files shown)


gengjiawen:
I think you missing my comment. This will still get npm 7. Because https://github.com/gitpod-io/workspace-images/issues/363#issuecomment-792407702. If you want npm 6, you should remove the update npm part.

https://github.com/gitpod-io/workspace-images/blob/0d86f18fcb06832838d2cb77636d2c1a6a7dbe6c/full/Dockerfile#L142

jankeromnes:
@gengjiawen indeed, you're right, sorry for misunderstanding your comment:

```
$ docker run --rm -it gitpod/workspace-full:branch-jx-upgrade-nvm /bin/bash

gitpod ~ $ npm --version
7.6.3
```

gengjiawen:
If you want bundled npm 6. Just remove npm after -g in this line `npm install -g npm typescript yarn`

jankeromnes:
> If you want bundled npm 6. Just remove npm after -g in this line `npm install -g npm typescript yarn`

Many thanks! Yes, that was my plan when I opened a workspace for this PR, but then I forgot about the tab and the workspace timed out again 😅 

Screenshot 2021-03-18 at 15 14 21

Anyway, let's try this again.

PR #376, opened 2021-03-25
[full] Update Composer to v2

caycehouse:
I have updated the way composer was installed, instead of using the version from apt which is still 1.0 I have found that by copying the file from the official composer docker image you can always get the latest stable version of composer put right in place. This fixes issue #339.

Solution found here for reference:
https://github.com/docker-library/docs/tree/master/composer#troubleshooting

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -1,5 +1,6 @@
 # Dazzle do not supports ARG TRIGGER_REBUILD=2
 FROM gitpod/workspace-base:latest
+FROM composer:2 AS composer
 
 RUN echo "ws full starts"
 
@@ -29,7 +30,6 @@ RUN install-packages \
         apache2 \
         nginx \
         nginx-extras \
-        composer \
         php \
         php-all-dev \
         php-bcmath \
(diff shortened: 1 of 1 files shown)


aledbf:
@caycehouse thank you for the contribution.

This is missing the `FROM` part, like `FROM composer:2 AS composer` at the top of the file

aledbf:
@caycehouse please squash the commits

jankeromnes:
Thanks! I'll trigger the CI [like so](https://github.com/gitpod-io/workspace-images/blob/master/CONTRIBUTING.md#how-to-make-circleci-build-a-pr) so that we can test the resulting images.

full/Dockerfile
```diff
@@ -1,5 +1,6 @@
 # Dazzle do not supports ARG TRIGGER_REBUILD=2
 FROM gitpod/workspace-base:latest
+FROM composer:2 AS composer
```
jankeromnes, line 3: **Thought**: Shouldn't these two lines be swapped? 🤔 (We want all the instructions below to happen in `gitpod/workspace-base:latest`, not in `composer:2`, except for the `COPY` but that already references `--from=composer` explicitly).

jankeromnes:
Ah, the build failed with

```
dazzle does not support multi-stage builds
```

To fix this, I guess we should either:

- Stop building `workspace-full` with Dazzle (conflicts with Gitpod's upcoming "on-the-fly image-layer-composition" feature)
- Make Dazzle accept multi-stage builds
- Not use multi-stage builds in `workspace-full`

strausmann:
When can this PR be completed? We have been waiting for the update of the PHP Composer to version 2 for a while...

princerachit:
Closing this as the composer changes are already shipped in https://github.com/gitpod-io/workspace-images/pull/660

PR #381, opened 2021-04-06
[full] add brew path first and level-up brew build step

gengjiawen:
* brew is more commonly used, so I moved to top part before php and node build staff
* brew has the latest toolchain like `curl`, `sqlite3` and other commonly used software. So I add `/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin/` before PATH, so we can use it first if brew installed related software.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -3,6 +3,20 @@ FROM gitpod/workspace-base:latest
 
 RUN echo "ws full starts"
 
+### Homebrew ###
+LABEL dazzle/layer=tool-brew
+LABEL dazzle/test=tests/tool-brew.yaml
+USER gitpod
+# Dazzle does not rebuild a layer until one of its lines are changed. Increase this counter to rebuild this layer.
+ENV TRIGGER_BREW_REBUILD=1
+RUN mkdir ~/.cache && /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+ENV PATH="/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin/:$PATH" \
+    MANPATH="$MANPATH:/home/linuxbrew/.linuxbrew/share/man" \
+    INFOPATH="$INFOPATH:/home/linuxbrew/.linuxbrew/share/info" \
+    HOMEBREW_NO_AUTO_UPDATE=1
+RUN sudo apt remove -y cmake \
+    && brew install cmake
+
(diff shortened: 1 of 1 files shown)


gengjiawen:
@jankeromnes PTAL.

jankeromnes:
Many thanks! Hmm, I vaguely remember brew-in-front-of-path breaking some things, but I don't remember what. Should be well tested I guess. 😅 Will try to review soon.

gengjiawen:
@jankeromnes Friendly ping.

gengjiawen:
Any progress ?

jankeromnes:
I think the problem with brew-in-front-of-path was that it overshadowed the system's libc with another one that broke tools.

Will trigger a CI run so that we can test the resulting image.

gengjiawen:
> I think the problem with brew-in-front-of-path was that it overshadowed the system's libc with another one that broke some tools.
> 
> Will trigger a CI run so that we can test the resulting image.

Looks our CI passed.

gengjiawen:
ping @jankeromnes

jankeromnes (commented):
Hi @gengjiawen! Sorry that we took so long to review your PR.

It looks good to me, and we can indeed move the entire HomeBrew Dazzle layer up a few levels (although I think the order doesn't have much of an influence with Dazzle builds).

However, please don't put HomeBrew first in the `$PATH`. Here are a few reasons why I don't think this is a good idea:
- HomeBrew is sort of "alien", especially on Linux (where its support is still somewhat new). I would rather prefer we give top priority to the "native" package managers like `apt-get`, where things are less likely to go wrong.
- We originally added HomeBrew so that you could `brew install X` without requiring `sudo`. However, now that all Gitpod users have `sudo`, they can also `sudo apt-get install X`, so having `brew` is less important today.
- Initially, we added HomeBrew first in the `$PATH`, and this caused some major problems for some projects because HomeBrew's `libc6` was different to Ubuntu's (and for some reason, the one from Ubuntu was needed). In case of duplicate packages, I'd prefer leaving priority to the ones from Ubuntu.

Please let me know if you disagree. 🙂

Otherwise, please simply rebase to fix the conflicts (or open a new Pull Request if that's easier), without putting HomeBrew first in the `$PATH`, and then we should be good to merge. ✅ 

Thanks again for improving Gitpod's workspace images! 💯 🙏

gengjiawen:
Thanks for the explanation. 

However, brew has advantages for it has newer software like sqlite, curl and git. The path issue will it work if you install them from brew.

But in general, I think you are right. It may cause more issues than it solve.

PR #390, opened 2021-04-26
[full] feat: Add rustfmt component for Rust setup

axonasif:
While working with rust over Gitpod you would always notice that the formatter is missing for `.rs` files(aka Rust project files) and you have to manually install it every-time. Although if I'm not wrong then Gitpod is meant to be bare-bones when it comes to ready to use workspaces ;) And since the additional formatter is being installed for [clang](https://github.com/gitpod-io/workspace-images/blob/409e1707b3cbe6752a894893cd6354fc13d7b8f5/full/Dockerfile#L18) and [golang](https://github.com/gitpod-io/workspace-images/blob/409e1707b3cbe6752a894893cd6354fc13d7b8f5/full/Dockerfile#L100) as well, why not for rustlang :smile: ?

Anyway, `rustfmt` comes by default unless you specify [minimal](https://github.com/gitpod-io/workspace-images/blob/409e1707b3cbe6752a894893cd6354fc13d7b8f5/full/Dockerfile#L196) profile for installing RustLang thru rustup.sh but that's totally fine, we can just install the formatter separately as done for clang and golang in the `full` workspace-image.

Thanks.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -198,6 +198,7 @@ RUN cp /home/gitpod/.profile /home/gitpod/.profile_orig && \
         rls \
         rust-analysis \
         rust-src \
+        rustfmt \
     && .cargo/bin/rustup completions bash | sudo tee /etc/bash_completion.d/rustup.bash-completion > /dev/null \
     && .cargo/bin/rustup completions bash cargo | sudo tee /etc/bash_completion.d/rustup.cargo-bash-completion > /dev/null \
     && grep -v -F -x -f /home/gitpod/.profile_orig /home/gitpod/.profile > /home/gitpod/.bashrc.d/80-rust


jankeromnes (approved):
Many thanks for this contribution @axonasif! 💯

> And since the additional formatter is being installed for clang and golang as well, why not for rustlang 😄 ?

That's an excellent point. 🎯 I think we should either include the formatters for these three languages, or remove them consistently.

So, let's re-install `rustfmt` now, and if the image becomes too big again, we may consider removing all formatters as one possible space-saving option.

Tested by opening this PR in Gitpod, and running:

```
$ rustfmt --version
error: 'rustfmt' is not installed for the toolchain '1.51.0-x86_64-unknown-linux-gnu'
To install, run `rustup component add rustfmt`

$ docker run -it --rm gitpod/workspace-full:branch-pr-390 /bin/bash
[...]
Status: Downloaded newer image for gitpod/workspace-full:branch-pr-390
gitpod ~ $ rustfmt --version
rustfmt 1.4.36-stable (7de6968 2021-02-07)
```

PR #392, opened 2021-04-28
Update Ruby to 2.7.3 in full image

tnir:
While #213 upgraded Ruby to 2.7 once, but reverted it in e9281a207c4c6b4c7df2e91e9ec81f36ed0652ae, Ruby 2.6.6 has some security vulnerabilities described in https://www.ruby-lang.org/en/news/2021/04/05/ruby-2-6-7-released/.
As Ruby 3.0.1 is also available, we can add it into the full image.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -180,8 +180,9 @@ RUN curl -fsSL https://rvm.io/mpapis.asc | gpg --import - \
     && curl -fsSL https://get.rvm.io | bash -s stable \
     && bash -lc " \
         rvm requirements \
-        && rvm install 2.6.6 \
-        && rvm use 2.6.6 --default \
+        && rvm install 3.0.1 \
+        && rvm install 2.6.7 \
+        && rvm use 2.6.7 --default \
         && rvm rubygems current \
         && gem install bundler --no-document \
         && gem install solargraph --no-document" \
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -180,9 +180,8 @@ RUN curl -fsSL https://rvm.io/mpapis.asc | gpg --import - \
     && curl -fsSL https://get.rvm.io | bash -s stable \
     && bash -lc " \
         rvm requirements \
-        && rvm install 3.0.1 \
-        && rvm install 2.6.7 \
-        && rvm use 2.6.7 --default \
+        && rvm install 2.7.3 \
+        && rvm use 2.7.3 --default \
         && rvm rubygems current \
         && gem install bundler --no-document \
         && gem install solargraph --no-document" \


tnir:
@jankeromnes What do you think?

jankeromnes:
> @jankeromnes What do you think?

Excellent catch! Many thanks @tnir.

However, in the interest of keeping `workspace-full` small, I'd prefer to only install a single version, so we should decide between `2.6.7` and `3.0.1`.

Meanwhile, I'll trigger a CI run so that we can test the image with both versions on actual repositories.

jankeromnes:
The build seems to have failed with:

```
Error running 'env GEM_HOME=/home/gitpod/.rvm/gems/ruby-2.6.7@global GEM_PATH= /home/gitpod/.rvm/rubies/ruby-2.6.7/bin/ruby -d /home/gitpod/.rvm/src/rubygems-3.0.9/setup.rb --no-document',
              please read /home/gitpod/.rvm/log/1620467082_ruby-2.6.7/rubygems.install.log
```

(Maybe that's a sign that we shouldn't install `2.6.7` and just go with `3.0.1`! 😅)

jankeromnes:
Tested with: https://gitpod.io/#https://github.com/jankeromnes/rails_sample_app

Unfortunately, this fails with:

```
```

Options:
- Bump `Gemfile.lock` (and any follow-up errors)
- Stick with Ruby `2.6.7` for now

jankeromnes:
Ruby 2.7.3 seems to work well with: https://gitpod.io/#https://github.com/jankeromnes/rails_sample_app

However, I couldn't get Ruby language smartness to work in VSCode, even after installing `rebornix.Ruby` (recommended) and reloading my browser.

jankeromnes (approved):
I can't get language smartness to work with `2.6.6` either, so this isn't a new problem (and might actually exist between keyboard and chair) so let's go with `2.7.3`. 🚀

Thanks again @tnir!

tnir:
Thank you, @jankeromnes!

What criteria do you have to merge a PR?

tnir:
> Tested with: https://gitpod.io/#https://github.com/jankeromnes/rails_sample_app
> 
> Unfortunately, this fails with:

> ruby_dep-1.5.0 requires ruby version >= 2.2.5, ~> 2.2, which is incompatible

https://github.com/gitpod-io/rails_sample_app/pull/3 and https://github.com/gitpod-io/rails_sample_app/pull/4 were pushed. Also see gitpod-io/gitpod#4243!

jankeromnes:
> What criteria do you have to merge a PR?

This is somewhat unofficial, but for PRs to workspace-images, the resulting images should:

1. Build successfully on CircleCI (we don't want to break the build)

2. Remain reasonably light (we want to keep `gitpod/workspace-full` as light as possible, e.g. by not installing large tools/packages that aren't useful to all Gitpod users -- or by installing & configuring large+complicated+popular packages as separate images, e.g. `gitpod/workspace-flutter`)

3. Ideally not break any downstream projects (e.g. we don't want to ship an upgrade that will suddenly break people's projects in Gitpod, or break Gitpod's own example projects, so we need to test the resulting images carefully and avoid any breaking changes when possible)

> gitpod-io/rails_sample_app#3 and gitpod-io/rails_sample_app#4 were pushed. Also see gitpod-io/gitpod#4243!

That's awesome! Many thanks for these fixes and the call-out. 🚀 This is very much appreciated. I'll advocate for these within the Gitpod team to create awareness. 😁

PR #419, opened 2021-06-11
[full] Add golangci-lint

rl-gitpod:
golangci-lint is the defacto go linter. Installing it in full will make it available for everyone to use. 
Worth noting that VSCode will fail silently if your config specifies golangci and it isn't present :-(

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -108,8 +108,8 @@ RUN curl -fsSL https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.t
         github.com/godoctor/godoctor && \
     GO111MODULE=on go get -v \
         golang.org/x/tools/gopls@v0.6.11 && \
-    sudo rm -rf $GOPATH/src $GOPATH/pkg /home/gitpod/.cache/go /home/gitpod/.cache/go-build && \
-    	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.40.1
+    curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.40.1 && \
+    sudo rm -rf $GOPATH/src $GOPATH/pkg /home/gitpod/.cache/go /home/gitpod/.cache/go-build
 # user Go packages
 ENV GOPATH=/workspace/go \
     PATH=/workspace/go/bin:$PATH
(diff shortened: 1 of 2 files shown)


full/Dockerfile
```diff
@@ -108,7 +108,8 @@ RUN curl -fsSL https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.t
         github.com/godoctor/godoctor && \
     GO111MODULE=on go get -v \
         golang.org/x/tools/gopls@v0.6.11 && \
-    sudo rm -rf $GOPATH/src $GOPATH/pkg /home/gitpod/.cache/go /home/gitpod/.cache/go-build
+    sudo rm -rf $GOPATH/src $GOPATH/pkg /home/gitpod/.cache/go /home/gitpod/.cache/go-build && \
+    	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.40.1
```
jankeromnes, line 112: Minor suggestions:
- Please fix indentation
- Maybe keep the final cleanup line as the last line?

```suggestion
    curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.40.1 && \
    sudo rm -rf $GOPATH/src $GOPATH/pkg /home/gitpod/.cache/go /home/gitpod/.cache/go-build
```

jankeromnes (commented):
Excellent idea, thanks @rl-gitpod! This makes a lot of sense to me.

Very minor suggestions in-line (feel free to address or dismiss) and a few questions:
- What should be done with `golang.org/x/lint/golint` in the `go get` list above?
- Is the above list even still relevant? (The top comment points to an old commit of vscode-go -- I believe the new list is here https://github.com/golang/vscode-go/blob/0f4b38ff0c60b9ac6c82bef484d78cad040e30fc/src/goTools.ts#L139-L155 but a bunch of tools are now installed conditionally)

rl-gitpod:
> Excellent idea, thanks @rl-gitpod! This makes a lot of sense to me.
> 
> Very minor suggestions in-line (feel free to address or dismiss) and a few questions:
> 
> * What should be done with `golang.org/x/lint/golint` in the `go get` list above?

I didn't want to remove anything in case it broke an existing workflow for someone using this image, but possibly being too defensive.
More generally it could be useful if we could indicate deprecated packages/setups in some way.

> * Is the above list even still relevant? (The top comment points to an old commit of vscode-go -- I believe the new list is here https://github.com/golang/vscode-go/blob/0f4b38ff0c60b9ac6c82bef484d78cad040e30fc/src/goTools.ts#L139-L155 but a bunch of tools are now installed conditionally)

Interesting.
I missed the significance of that - there is a significant drift!

I'll try out the newer list.

full/Dockerfile
rl-gitpod: Thanks!!! I thought I had fixed the indentation:-(

jankeromnes:
Hi @rl-gitpod! What's the status of this PR? 😇 (I.e. are you still working on it, or are you awaiting a review?)

I think once this is done, we should point a few forks to workspace images from this PR (e.g. `gitpod/workspace-full:branch-rl-gitpod-add-golangci-lint-to-418`) and then confirm that everything is still working as intended.

aledbf (approved):
LGTM

Issue #441, opened 2021-07-07
gitpod/workspace-full-vnc:5900 is 502


ghuntley:
### Bug description

<img width="922" alt="2021-07-06_22-38-05" src="https://user-images.githubusercontent.com/127353/124601155-d7b98b80-deaa-11eb-8bc3-a56905e3ce54.png">

<img width="1054" alt="2021-07-06_22-37-48" src="https://user-images.githubusercontent.com/127353/124601129-cec8ba00-deaa-11eb-8111-55a163a729a4.png">

### Steps to reproduce

1. https://gitpod.io/#https://github.com/ghuntley/gitpod-vnc
2. Open port 5900

### Expected behavior

VNC session!

### Example repository

https://github.com/ghuntley/gitpod-vnc

ghuntley:
Using this monkeypatch https://github.com/ghuntley/gitpod-vnc/blob/64b372fbc638d0ce4854cdf0d8ac0df4a64881de/.gitpod.yml#L4 the VNC server now starts.

```
gitpod /workspace/gitpod-vnc $ ps aux |grep vnc
gitpod        33  9.3  0.1 932664 82380 ?        RNl  04:32   0:02 /ide/node/bin/gitpod-node ./out/gitpod.js /workspace/gitpod-vnc --port 23000 --hostname 0.0.0.0
gitpod       811  0.0  0.0  40136  8988 ?        SNs  04:32   0:00 x11vnc -localhost -shared -display :0 -forever -rfbport 5900 -bg -o /tmp/x11vnc-0.log
gitpod       812  0.0  0.0      0     0 pts/0    ZN   04:32   0:00 [start-vnc-sessi] <defunct>
gitpod      1015  0.0  0.0   6312   680 pts/0    SN+  04:33   0:00 grep --color=auto vnc
gitpod /workspace/gitpod-vnc $ 
```

```
gitpod /workspace/gitpod-vnc $ cat /tmp/x11vnc-0.log 
07/07/2021 04:32:41 passing arg to libvncserver: -rfbport
07/07/2021 04:32:41 passing arg to libvncserver: 5900
###############################################################
#@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@#
#@                                                           @#
#@  **  WARNING  **  WARNING  **  WARNING  **  WARNING  **   @#
#@                                                           @#
[...]

PR #442, opened 2021-07-07
[full-vnc] Fix Virtual Desktop (rename noVNC/utils/launch.sh → noVNC/utils/novnc_proxy)

jankeromnes:
Credit to @ghuntley for identifying the root cause: https://github.com/gitpod-io/workspace-images/issues/441#issuecomment-875249357

Fixes https://github.com/gitpod-io/workspace-images/issues/441

Fixes https://github.com/gitpod-io/gitpod/issues/4770


csweichel (approved):
LGTM

jankeromnes:
Well, I didn't fully test this, but using the correct script name can not do more harm than using the wrong one. So, merging as is and iterating further if needed.

PR #443, opened 2021-07-07
full/Dockerfile: Install ripgrep

Panchajanya1999:
Ripgrep is a very important and easy to use/handy for searching strings. 
Installing it everytime is PITA.

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -18,7 +18,7 @@ RUN curl -o /var/lib/apt/dazzle-marks/llvm.gpg -fsSL https://apt.llvm.org/llvm-s
         clang-format \
         clang-tidy \
         gdb \
-        lld
+        lld ripgrep
 
 ### Apache, PHP and Nginx ###
 LABEL dazzle/layer=tool-nginx


jankeromnes:
Hi @Panchajanya1999! Many thanks for this contribution. I agree that ripgrep is an awesome tool. 💯

In fact, did you know that the IDE code search feature is based on ripgrep? (`rg` is not currently installed in the Terminal, but it's installed in the IDE image, which explains why the code search feature is so scary-fast 😁.)

I agree that having it by default in the Terminal would be also great. ⚡

full/Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN curl -o /var/lib/apt/dazzle-marks/llvm.gpg -fsSL https://apt.llvm.org/llvm-s
         clang-format \
         clang-tidy \
         gdb \
-        lld
+        lld ripgrep
```
jankeromnes, line 21: Semantically, I'm not sure the C/C++ section is the most appropriate to install `ripgrep` in. Maybe it should be installed in `base/Dockerfile` instead?

Also, please at least keep one package per line, and sort packages alphabetically. 😇

Panchajanya1999: Ah well. I am setting up in base/Dockerfile instead

PR #447, opened 2021-07-14
Display current git branch in the terminal prompt (PS1)

Shulammite-Aso:
Fixes https://github.com/gitpod-io/gitpod/issues/3284

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -43,7 +43,7 @@ RUN useradd -l -u 33333 -G sudo -md /home/gitpod -s /bin/bash -p gitpod gitpod \
 ENV HOME=/home/gitpod
 WORKDIR $HOME
 # custom Bash prompt
-RUN { echo && echo "PS1='\[\e]0;\u \w\a\]\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w\[\033[00m\] \\\$ '" ; } >> .bashrc
+RUN { echo && echo "PS1='\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w\[\033[00m\]$(__git_ps1 " (%s)") $ '" ; } >> .bashrc
 
 ### Gitpod user (2) ###
 USER gitpod
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -43,7 +43,7 @@ RUN useradd -l -u 33333 -G sudo -md /home/gitpod -s /bin/bash -p gitpod gitpod \
 ENV HOME=/home/gitpod
 WORKDIR $HOME
 # custom Bash prompt
-RUN { echo && echo "PS1='\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w\[\033[00m\]$(__git_ps1 " (%s)") $ '" ; } >> .bashrc
+RUN { echo && echo "PS1='\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w\[\033[00m\]\$(__git_ps1 \" (%s)\") $ '" ; } >> .bashrc
 
 ### Gitpod user (2) ###
 USER gitpod


Shulammite-Aso:
@jankeromnes, please take a look.

jankeromnes:
Super-cool! Many thanks @Shulammite-Aso for taking on & fixing this issue 💯 x 🙏 

Also, sorry for the delay in reviewing this! I'll take a look now, and I'll also make CircleCI build your PR so that we get images we can test on actual repositories. 😊

Shulammite-Aso:
Alrigth @jankeromnes, can't wait! :smiley:

base/Dockerfile
```diff
@@ -43,7 +43,7 @@ RUN useradd -l -u 33333 -G sudo -md /home/gitpod -s /bin/bash -p gitpod gitpod \
 ENV HOME=/home/gitpod
 WORKDIR $HOME
 # custom Bash prompt
-RUN { echo && echo "PS1='\[\e]0;\u \w\a\]\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w\[\033[00m\] \\\$ '" ; } >> .bashrc
+RUN { echo && echo "PS1='\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w\[\033[00m\]$(__git_ps1 " (%s)") $ '" ; } >> .bashrc
```
jankeromnes, line 46: I'd like to share two thoughts here:

1. I notice that you are removing the `\[\e]0;\u \w\a\]` part of the `PS1` prompt. I initially had no idea what this was doing (and when testing it in Gitpod, it seemed to have no effect), but after a while I found out that this apparently sets the Terminal's _title bar_ ([source](https://unix.stackexchange.com/a/306721)). I think setting the Terminal title that way used to work in Gitpod's previous IDE (Theia), but it no longer works in the new IDE (VS Code), so it should be safe to remove. 👍 

2. While testing the proposed new syntax, the `$(__git_ps1 " (%s)")` part gave me surprising results: The branch name initially seemed correct, but it didn't update when changing directories or when switching to different Git branches (in the screenshot below, I've copy/pasted the above `RUN` line into the left Terminal, and then opened a new Terminal on the right in order to test the prompt -- as you'll notice, the branch name always stayed the same, even outside the Git repository or on a different branch):

Screenshot 2021-07-19 at 13 30 37

I believe this is because the special `$` character needs additional "escaping" here (i.e. it needs some extra `\` character(s) in front of it).

Since it's quite awkward and difficult to write a correct PS1 syntax in this situation (i.e. in a Dockerfile, that prints a command line into a `.bashrc` script file, that then gets interpreted by IDE Terminals 🤯), I believe the correctly escaped syntax would be `\$(__git_ps1 \" (%s)\")`, i.e. like so:

```suggestion
RUN { echo && echo "PS1='\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w\[\033[00m\]\$(__git_ps1 \" (%s)\") $ '" ; } >> .bashrc
```

jankeromnes (commented):
Thanks again for this great work @Shulammite-Aso 😁

I believe it's almost perfect, but noticed a very small issue, that basically just requires adding a few `\` to the correct places.

Please simply add the missing `\` where expected (see the code suggestion below), and then I'll trigger a new CircleCI build and we'll be able to test the updated fix in Gitpod.

I've also added some more thoughts & details (sorry for being so verbose 😅) but feel free to skip them if you like. (Or, if you have any question / if anything seems unclear, please let me know and I'll be happy to help!)

Shulammite-Aso:
Okay @jankeromnes. Taking a moment to read, understand, then make the necessary changes. Will let you know if I have a question.

Shulammite-Aso:
@jankeromnes, I get this now and trying to make the changes, but I just saw that I have reached my monthly workspace hour limit, sadly.  Now I don't know if there's any other way around this?  I've been playing around using the workspaces for the past weeks. It almost slipped my mind that I'm on a very limited plan. 🤦‍♀️

Shulammite-Aso:
commited!

> Given that you're contributing to Gitpod, I've just granted you 20 bonus hours,

Thanks for this. Will start keeping my limit in mind.

jankeromnes:
Alright, many thanks for the fix! This now looks perfect. ✨

I've also made CircleCI build this Pull Request, which resulted in the Docker image `gitpod/workspace-full:build-branch-pr-447` getting published to Docker Hub. Let's hook this up with a repository, and briefly test it in Gitpod, to confirm that everything works as intended.

jankeromnes (approved):
Alright, this works like a charm! 🎉 🎆 Thus I'm approving this Pull Request, and it can be merged soon.

Screenshot 2021-07-20 at 09 35 49

Thanks again for implementing this!

---

I've just noticed one tiny problem, but it seems completely unrelated to this code change (probably a random new bug in our code): Every new IDE Terminal now shows this error as the first line:

```
ls: cannot access '/home/gitpod/.bashrc.d/*': No such file or directory
```

I'll investigate this a bit more, to see if I can find a fix for it, but otherwise I think we should merge your PR quite soon. 😊

Shulammite-Aso:
> Alright, this works like a charm! :tada: :fireworks: Thus I'm approving this Pull Request, and it can be merged soon.

nice to see!! :woman_dancing:  :tada: :tada:

Shulammite-Aso:
Thanks for letting me contribute @jankeromnes!! :grin: :grin:

jankeromnes:
> ```
> ls: cannot access '/home/gitpod/.bashrc.d/*': No such file or directory
> ```
> 
> I'll investigate this a bit more, to see if I can find a fix for it, but otherwise I think we should merge your PR quite soon. 😊

Ah, I think I understand the problem now -- to test this Pull Request, I've used the `gitpod/workspace-base` image instead of the more complete `gitpod/workspace-full`. It would seem that this line logs an error when the `$HOME/.bashrc.d` directory is empty (while instead it should not log an error and just do nothing):

https://github.com/gitpod-io/workspace-images/blob/d57c5f4b47a2df0bad344ff6c6a0209bd862a352/base/Dockerfile#L54

In other words, this bug existed before, and has indeed nothing to do with this Pull Request -- so I'll merge this now, and fix the bug in a separate Pull Request. 😁 🙌

> Thanks for letting me contribute @jankeromnes!! 😁 😁

You're quite welcome! 😁 And all the thanks go to you, for doing all the work 🙏

Shulammite-Aso:
>You're quite welcome! :grin: And all the thanks go to you, for doing all the work :pray:

:tada::tada::tada:

@jankeromnes I'm also willing to make another contribution!  I know I should probably not take up another `good-first-issue`, as it will no longer be my first. Unless it's absolutely okay to do that.:smile:   Otherwise I can make changes to a react/typescript and nodejs/typescript codebase. I'm yet to comfortably pick up Golang, hopefully, I do so soon.

ArthurSens:
Hey @jankeromnes, is this deployed alongside Gitpod when we update production? I'm not sure if I'm supposed to move this to "Done" in the groundwork project 🤔

ArthurSens:
Friendly ping for the question above /\

PR #450, opened 2021-07-19
feat(base): add git-lfs

ghuntley:
Adds `git-lfs` and installs the shell hooks at the system level (which can be overridden by configuring at the user scope).

Related to https://github.com/gitpod-io/gitpod/issues/4852

Tested with

```cd base; docker build -f Dockerfile .```


aledbf (approved):
LGTM

PR #458, opened 2021-08-01
Update rust to 1.54.0, node to v14.17.4 and docker-compose to 1.29.2

aledbf:

PR #461, opened 2021-08-01
[python] Set PIP_USER=no

csweichel:
fixes https://github.com/gitpod-io/gitpod/issues/4886

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -154,6 +154,7 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
         mypy autopep8 pep8 pylama pydocstyle bandit notebook \
         twine \
     && sudo rm -rf /tmp/*
+ENV PIP_USER=no
 # Gitpod will automatically add user site under `/workspace` to persist your packages.
 # ENV PYTHONUSERBASE=/workspace/.pip-modules \
 #    PIP_USER=yes
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -155,9 +155,9 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
         twine \
     && sudo rm -rf /tmp/*
 ENV PIP_USER=no
-# Gitpod will automatically add user site under `/workspace` to persist your packages.
-# ENV PYTHONUSERBASE=/workspace/.pip-modules \
-#    PIP_USER=yes
+ENV PIPENV_VENV_IN_PROJECT=true
+ENV PYTHONUSERBASE=/workspace/.pip-modules
+ENV PATH=$PYTHONUSERBASE/bin:$PATH
 
 ### Ruby ###
 LABEL dazzle/layer=lang-ruby
@@ -176,6 +176,10 @@ RUN curl -fsSL https://rvm.io/mpapis.asc | gpg --import - \
     && echo '[[ -s "$HOME/.rvm/scripts/rvm" ]] && source "$HOME/.rvm/scripts/rvm" # Load RVM into a shell session *as a function*' >> /home/gitpod/.bashrc.d/70-ruby
 RUN echo "rvm_gems_path=/workspace/.rvm" > ~/.rvmrc
 
+ENV GEM_HOME=/workspace/.rvm
+ENV GEM_PATH=$GEM_HOME:$GEM_PATH
+ENV PATH=/workspace/.rvm/bin:$PATH
+
 ### Rust ###
 LABEL dazzle/layer=lang-rust
 LABEL dazzle/test=tests/lang-rust.yaml
@@ -191,6 +195,12 @@ RUN cp /home/gitpod/.profile /home/gitpod/.profile_orig && \
     && .cargo/bin/rustup completions bash cargo | sudo tee /etc/bash_completion.d/rustup.cargo-bash-completion > /dev/null \
     && grep -v -F -x -f /home/gitpod/.profile_orig /home/gitpod/.profile > /home/gitpod/.bashrc.d/80-rust
 ENV PATH=$PATH:$HOME/.cargo/bin
+# TODO: setting CARGO_HOME to /workspace/.cargo avoids manual updates. Remove after full workspace backups are GA.
+ENV CARGO_HOME=/workspace/.cargo
+ENV PATH=$CARGO_HOME/bin:$PATH
+
+RUN sudo mkdir -p $CARGO_HOME \
+    && sudo chown -R gitpod:gitpod $CARGO_HOME
 
 RUN bash -lc "cargo install cargo-watch cargo-edit cargo-tree cargo-workspaces"
 


aledbf (approved):
LGTM

aledbf:
@csweichel let's merge https://github.com/gitpod-io/workspace-images/pull/458 before this one, cancel the circle CI build, and merge this PR to release the new images?

jankeromnes:
@csweichel @aledbf Many thanks! 💯

@AlexTugarev Since you previously worked on the Python setup, please have a look as well if you have time. 🙂

Also, on a more general note, I believe we should soon do two things with our Python setup:

- Drop support for (deprecated) Python 2, and in the same run maybe remove Pyenv entirely (and keep only Ubuntu's stock Python) -- relatedly, now that we have Full Workspace Backup generally enabled, I think we can remove a ton of `/workspace` hacks

- Compare our Python stack with https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/python-debian.sh and see if maybe there are a few things we could do better (e.g. some additional tools to install, or if there exists a more common / saner Python dev setup)

full/Dockerfile
```diff
@@ -154,6 +154,7 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
         mypy autopep8 pep8 pylama pydocstyle bandit notebook \
         twine \
     && sudo rm -rf /tmp/*
+ENV PIP_USER=no
```
AlexTugarev, line 157: why unsetting it here whereas its set in [gitpod-layer](https://github.com/gitpod-io/gitpod/blob/24f7b609bf79b247d72d1841282639726bcd5891/components/image-builder/workspace-image-layer/gitpod-layer/alpine/gitpod/.bashrc-append#L11) on top? my first impression is that this won't be effective. 

also, just a reminder, without forcibly enabling `PYTHONUSERBASE` under `/workspace` the backup strategy for python is gone, and prebuilds as well. using a virutal env setup seems to be the best way. it would be great to provide a replacement that works.

AlexTugarev:
> ... now that we have Full Workspace Backup generally enabled ...

@jankeromnes, we do?

jankeromnes:
> > ... now that we have Full Workspace Backup generally enabled ...
> 
> @jankeromnes, we do?

We don't? 😳 @csweichel @aledbf do you know?

aledbf:
> We don't? flushed @csweichel @aledbf do you know?

It is not enabled yet (there is an issue with the user permissions when we create the workspace backup)

jankeromnes (changes requested):
Thanks again for improving Gitpod's Python setup! 💯

Blocking this PR until https://github.com/gitpod-io/workspace-images/pull/461#discussion_r680693077 is resolved, so that we don't accidentally merge it as is.

full/Dockerfile
aledbf: @AlexTugarev the gitpod-layer was removed https://github.com/gitpod-io/gitpod/pull/4923. I will add the `.bashrc-append` changes here.

aledbf:
@csweichel @jankeromnes @AlexTugarev can we merge this PR? Without it, most of the examples are broken when the new image builder mk3 is used (due to the removal of the gitpod layer)
(needs a rebase first)

full/Dockerfile
```diff
@@ -175,6 +176,10 @@ RUN curl -fsSL https://rvm.io/mpapis.asc | gpg --import - \
     && echo '[[ -s "$HOME/.rvm/scripts/rvm" ]] && source "$HOME/.rvm/scripts/rvm" # Load RVM into a shell session *as a function*' >> /home/gitpod/.bashrc.d/70-ruby
 RUN echo "rvm_gems_path=/workspace/.rvm" > ~/.rvmrc
 
+ENV GEM_HOME=/workspace/.rvm
+ENV GEM_PATH=$GEM_HOME:$GEM_PATH
+ENV PATH=/workspace/.rvm/bin:$PATH
```
jankeromnes, line 181: I think this does more harm than good. Maybe let's remove it for now?

```suggestion
```

full/Dockerfile
```diff
@@ -154,9 +154,10 @@ RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-ins
         mypy autopep8 pep8 pylama pydocstyle bandit notebook \
         twine \
     && sudo rm -rf /tmp/*
-# Gitpod will automatically add user site under `/workspace` to persist your packages.
-# ENV PYTHONUSERBASE=/workspace/.pip-modules \
-#    PIP_USER=yes
+ENV PIP_USER=no
+ENV PIPENV_VENV_IN_PROJECT=true
+ENV PYTHONUSERBASE=/workspace/.pip-modules
+ENV PATH=$PYTHONUSERBASE/bin:$PATH
```
jankeromnes, line 160: I have no Python experience to tell whether setting any of these variables is okay.

However, I don't think you'll find any reviewers here with Python experience, so continuing to wait for a review here is unhelpful.

Also, if possible, I would very much prefer not setting any of these variables. I'm aware that this may break package persistence (they'll land outside of `/workspace`) but for that I'd rather focus on enabling Full Workspace Backup instead.

full/Dockerfile
```diff
@@ -190,7 +195,14 @@ RUN cp /home/gitpod/.profile /home/gitpod/.profile_orig && \
...
     && grep -v -F -x -f /home/gitpod/.profile_orig /home/gitpod/.profile > /home/gitpod/.bashrc.d/80-rust
 ENV PATH=$PATH:$HOME/.cargo/bin
+# TODO: setting CARGO_HOME to /workspace/.cargo avoids manual updates. Remove after full workspace backups are GA.
+ENV CARGO_HOME=/workspace/.cargo
+ENV PATH=$CARGO_HOME/bin:$PATH
 
+USER root
+RUN mkdir -p $CARGO_HOME \
+    && chown -R gitpod.gitpod $CARGO_HOME
+USER gitpod
```
jankeromnes, line 205: Can be simplified. Also, shouldn't it be `gitpod:gitpod` instead of `gitpod.gitpod`?

```suggestion
RUN sudo mkdir -p $CARGO_HOME \
    && sudo chown -R gitpod:gitpod $CARGO_HOME
```

jankeromnes (approved):
Approving to unblock this PR, but I would very very much prefer it if we remove all of these silly variables.

- For Python, we're unwittingly breaking many people's setups by forcing some PIP_USER or VENV things

- For Ruby, our setup is so broken that every major Ruby project just replaces it entirely in a Dockerfile (also, we use RVM instead of today's standard Rbenv)

- For Rust, the /workspace hack still looks okay, but here too I'd very much prefer removing the hack altogether, and instead enable Full Workspace Backup everywhere

full/Dockerfile
aledbf: done

full/Dockerfile
aledbf: >  but for that I'd rather focus on enabling Full Workspace Backup instead.

this is one of those issues where we have a circular dependency; we need this PR to switch to the new image builder mk3. With that, we can focus on the dazzle refactor that will reduce the image size and simplify the dependencies tangle, and then we can start using stargz. After that, we should focus on Full Workspace Backup. Without all those previous steps we are not improving the size, usability, or modularity of the workspace container images.

PR #479, opened 2021-08-21
[full] move go get to go install and add precompile for golangci-lint

MedzikUser:
1. go get is deprecated
2. installing golangci-lint from source code isn't recommended


MedzikUser:
DO NOT MERGE

PR #482, opened 2021-08-26
feat(workspace-full): install common cloud CLI tooling

ghuntley:
Tested through:

1. docker build -f .
2. dazzle integration tests

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,6 +1,13 @@
 image: gitpod/workspace-full
 
 tasks:
+  - env:
+      # https://github.com/gitpod-io/gitpod/issues/479
+      PIP_USER: no
+  - init:
+      brew install pre-commit
+      pre-commit install
+      pre-commit run
   - command: docker
 
 vscode:
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,13 +1,6 @@
 image: gitpod/workspace-full
 
 tasks:
-  - env:
-      # https://github.com/gitpod-io/gitpod/issues/479
-      PIP_USER: no
-  - init:
-      brew install pre-commit
-      pre-commit install
-      pre-commit run
   - command: docker
 
 vscode:
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -237,6 +237,32 @@ RUN curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/focal.gpg | sudo apt-key
     && apt-get update \
     && apt-get install -y tailscale
 
+### Cloud Infrastructure CLIs ###
+
+LABEL dazzle/layer=tool-cloud-infra-cli
+LABEL dazzle/test=tests/tool-cloud-infra-cli.yaml
+USER root
+
+# Amazon Web Services
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" \
+    && unzip awscliv2.zip \
+    && ./aws/install \
+    && npm install -g aws-cdk
+
+# Digital Ocean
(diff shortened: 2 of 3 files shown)


full/Dockerfile
```diff
@@ -227,6 +227,32 @@ RUN curl -o /tmp/dive.deb -fsSL https://github.com/wagoodman/dive/releases/downl
     && apt install /tmp/dive.deb \
     && rm /tmp/dive.deb
 
+### Cloud Infrastructure CLIs ###
+
+LABEL dazzle/layer=tool-cloud-infra-cli
+LABEL dazzle/test=tests/tool-cloud-infra-cli.yaml
+USER root
+
+# Amazon Web Services
```
rl-gitpod, line 236: Consider having each on their own layer as they can change independently of each other?

Also version pinning might be worthwhile as I've been bitten by updates that break existing usages of aws-cli for e.g.

.gitpod.yml
```diff
@@ -1,6 +1,13 @@
 image: gitpod/workspace-full
 
 tasks:
+  - env:
+      # https://github.com/gitpod-io/gitpod/issues/479
+      PIP_USER: no
+  - init:
+      brew install pre-commit
```
rl-gitpod, line 8: This takes a good amount of time on every ws startup. Can it be installed as part of the image?

.gitpod.yml
```diff
@@ -1,6 +1,13 @@
...
+      pre-commit install
+      pre-commit run
```
rl-gitpod, line 10: This is failing for me:

```
$ pre-commit run
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Initializing environment for https://github.com/jumanjihouse/pre-commit-hooks.
[INFO] Initializing environment for https://github.com/igorshubovych/markdownlint-cli.
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('/home/gitpod/.cache/pre-commit/repopsdnf9zr/py_env-python3.9/bin/python', '-mpip', 'install', '.')
return code: 1
expected return code: 0
stdout: (none)
stderr:
    ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```

full/Dockerfile
ghuntley: ❓ Do we need to/is it worth the complexity?

💭 Usually would agree with pinning but as these tools interface with 3rd party vendors whos API's change on a whim I think it's best to run unpinned. If someone wants to pin then they should create their own Dockerimage (which we highly recommend) instead of consuming our defaults.

.gitpod.yml
ghuntley: I'll back it out of this pull-request and move 🆙 into workspace-full by default.

.gitpod.yml
ghuntley: The key is PIP_USER: no

```
# https://github.com/gitpod-io/gitpod/issues/479
PIP_USER: no
```

full/Dockerfile
rl-gitpod: > Do we need to/is it worth the complexity?

Pros and cons (as always), hence the *might be*.

> 3rd party vendors whos API's change on a whim...
>  consuming our defaults.

It is not a particularly stable default i.e. it is whatever was current for that particular vendor when the image was last updated for a likely unrelated reason.

I guess that is acceptable for the generic 'workhorse' image - *a* version of that tool is available, we just don't guarantee which. I don't know the use cases driving the adoption of this in workspace-full as opposed to say workspace-cloud-infra for e.g.

Depending on those use use cases, is it worth considering a check for major version number for each?
e.g. in the tests adding something like:

> - stdout.indexOf("aws-cli/2.") != -1

That way we can say our defaults are 'v2.x of AWS cli' which is somewhat more predictable and perhaps more importantly, documentable.

princerachit:
I am closing this PR as this is obsolete. Further, we do not want to increase size of full image.

PR #483, opened 2021-08-29
fix: check if .bashrc.d is empty on base image

### LINKED ISSUE #480: bashrc.d is empty on base image
When deriving an image from `base` you will get an error on the shell:

    ls: cannot access '/home/gitpod/.bashrc.d/*': No such file or directory

This happens because `base` has this:

https://github.com/gitpod-io/workspace-images/blob/14cf434489db719e93b62024cdea7ad557c70574/base/Dockerfile#L53-L56

But no files are created on that folder (`full` does).

Either `base` should create a dummy file there to avoid the error, or a prior test should be done to be sure there is something for `ls` to list.

progmatic-99:
Fixes #480

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -52,8 +52,8 @@ USER gitpod
 # use sudo so that user does not get sudo usage info on (the first) login
 RUN sudo echo "Running 'sudo' for Gitpod: success" && \
     # create .bashrc.d folder and source it in the bashrc
-    mkdir /home/gitpod/.bashrc.d && \
-    (echo; echo "for i in \$(ls \$HOME/.bashrc.d/*); do source \$i; done"; echo) >> /home/gitpod/.bashrc
+    mkdir -p /home/gitpod/.bashrc.d && \
+    (echo; echo "for i in \$(ls -A \$HOME/.bashrc.d/); do source \$HOME/.bashrc.d/\$i; done"; echo) >> /home/gitpod/.bashrc
 
 # configure git-lfs
 RUN sudo git lfs install --system


base/Dockerfile
```diff
@@ -52,8 +52,10 @@ USER gitpod
 # use sudo so that user does not get sudo usage info on (the first) login
 RUN sudo echo "Running 'sudo' for Gitpod: success" && \
     # create .bashrc.d folder and source it in the bashrc
-    mkdir /home/gitpod/.bashrc.d && \
-    (echo; echo "for i in \$(ls \$HOME/.bashrc.d/*); do source \$i; done"; echo) >> /home/gitpod/.bashrc
+    mkdir -p /home/gitpod/.bashrc.d && \
+    if [ "$(ls -A $HOME/.bashrc.d/)" ]; then
```
telenieko, line 56: At this stage `/home/gitpod/.bashrc.d` will always be empty when the check is done (the `if` statement is done at build time not runtime).

the check should be done in `/home/gitpod/.bashrc` (runtime). Thus, the if statement should be with the for loop.

telenieko:
> Fixes #480

Also Fixes https://github.com/gitpod-io/gitpod/issues/4044

base/Dockerfile
telenieko: Untested solution:

```
    (echo; echo "for i in \$(ls -A \$HOME/.bashrc.d/); do source \$HOME/.bashrc.d/\$i; done"; echo) >> /home/gitpod/.bashrc
```

The idea here is that `ls -A` on the directory itself comes empty if there are no files, and print the files (filename only) otherwise.

As `ls -A` prints only the filename the path needs amendment on the source line later.

base/Dockerfile
```diff
@@ -52,8 +52,8 @@ USER gitpod
 # use sudo so that user does not get sudo usage info on (the first) login
 RUN sudo echo "Running 'sudo' for Gitpod: success" && \
     # create .bashrc.d folder and source it in the bashrc
-    mkdir /home/gitpod/.bashrc.d && \
-    (echo; echo "for i in \$(ls \$HOME/.bashrc.d/*); do source \$i; done"; echo) >> /home/gitpod/.bashrc
+    mkdir -p /home/gitpod/.bashrc.d && \
+    (echo; echo -e "if ! [ -z \$(ls -A -- \$HOME/.bashrc.d/); then\nfor i in \$(ls -A \$HOME/.bashrc.d/); do source \$HOME/.bashrc.d/\$i; done\nfi"; echo) >> /home/gitpod/.bashrc
```
telenieko, line 56: The opening `[` of the if statement is unclosed? 

If using `ls -A` you can skip the if test and go directly to the for loop, because `ls -A` will come out empty if the folder is empty already. 

So you may simplify to:

```shell
(echo; echo "for i in \$(ls -A \$HOME/.bashrc.d/); do source \$HOME/.bashrc.d/\$i; done"; echo) >> /home/gitpod/.bashrc
```

Manual test:

```shell
docker run -it --rm gitpod/workspace-base

# remove the last two lines (the current failing code)
gitpod ~ $ sed -i '$ d'  /home/gitpod/.bashrc
gitpod ~ $ sed -i '$ d'  /home/gitpod/.bashrc

# Now add the new code:
gitpod ~ $ (echo; echo "for i in \$(ls -A \$HOME/.bashrc.d/); do source \$HOME/.bashrc.d/\$i; done"; echo) >> /home/gitpod/.bashrc

# verify result
gitpod ~ $ tail -2  /home/gitpod/.bashrc
for i in $(ls -A $HOME/.bashrc.d/); do source $HOME/.bashrc.d/$i; done

# enter a new shell, see no error appears now
gitpod ~ $ bash
gitpod ~ $ exit

# now create a file an see it is sourced
gitpod ~ $ echo "echo 'I was sourced'" > $HOME/.bashrc.d/10_test.sh
gitpod ~ $ bash
I was sourced
gitpod ~ $ exit
```



```

progmatic-99: Thanks for the comments.

telenieko:
Looks good to me. Let's see if @corneliusludmann or someone can verify

corneliusludmann (approved):
Thanks for the contribution! Works as advertised!

PR #486, opened 2021-08-31
Update full-vnc image to better support JavaFX dev

PEZ:
When developing JavaFX (and probably also AWT) apps, the contents of them is not visible. This change makes java processes start with an option that fixes (or does not provoke, maybe is a better description) the problem.

See https://github.com/gitpod-io/gitpod/issues/1004 for a discussion about the problem.

### DIFF
--- a/full-vnc/Dockerfile
+++ b/full-vnc/Dockerfile
@@ -9,6 +9,10 @@ ENV WINDOW_MANAGER="openbox"
 
 USER root
 
+# Make JavaFX display windows with visible content 
+RUN echo export JAVA_TOOL_OPTIONS=\"\$JAVA_TOOL_OPTIONS -Dsun.java2d.xrender=false\" >> /home/gitpod/.bashrc
+
+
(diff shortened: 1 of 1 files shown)


full-vnc/Dockerfile
```diff
@@ -9,6 +9,12 @@ ENV WINDOW_MANAGER="openbox"
 
 USER root
 
+# Make JavaFX display windows with visible content
+RUN echo >> /home/gitpod/.bashrc
```
gengjiawen, line 13: Can you make this in one-line to make the docker layer less.

PEZ: Oh, sorry for dropping the ball here. Totally forgot. On it!

PEZ:
Let me know if I should do something more here. Thanks!

PEZ:
Sorry for the -e error there. I am a Docker noob.

princerachit:
This is obsolete as we have moved to dazzlev2

PR #520, opened 2021-10-05
[Full]: Allow to persist global node_modules

artalar:
The changes is allowing user installing global node modules and keep it between workspace inactivity

The naming is come from https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally#manually-change-npms-default-directory

### DIFF
--- a/full/Dockerfile
+++ b/full/Dockerfile
@@ -125,10 +125,13 @@ LABEL dazzle/test=tests/lang-node.yaml
 USER gitpod
 ENV NODE_VERSION=14.17.6
 ENV TRIGGER_REBUILD=1
+ENV PATH=/workspace/.npm-global/bin:$PATH
 RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | PROFILE=/dev/null bash \
     && bash -c ". .nvm/nvm.sh \
         && nvm install $NODE_VERSION \
         && nvm alias default $NODE_VERSION \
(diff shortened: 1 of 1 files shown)


artalar:
@jankeromnes check this out< please

artalar:
@ghuntley check this out, please

kylos101:
Hello @artalar , what is your use case? For example, are you wishing to use global modules, without having to constantly reinstall them?  Or are you installing and removing and installing often (because you're testing global modules)?

If it's the former, I recommend a [custom docker image](https://www.gitpod.io/docs/config-docker#configure-a-custom-dockerfile), using `workspace-node` as a base, and you could then install the global modules as part of the custom docker image.

If it's the latter, let us know?

artalar:
Hi @kylos101 , thanks for response, sorry for my delay.
My special case is necessity of `lerna` global installation for some scripts.
I will rebase from master today.

kylos101:
Hi @artalar , okay, you might want to [try creating your own dockerfile in your own repo](https://www.gitpod.io/docs/config-docker#configure-a-custom-dockerfile), and setup global node_modules there. This way, you can own your customization, and it does not have to be in a Gitpod workspace base image (shared by others).

Given what you've shared, it sounds like the customization you desire is specific to you and your use case, rather than a contribution that is for base image sin general. 

I am going to close this PR for now. If you continue to have trouble, may I ask you to open an issue, describing why a custom image will not work?

PR #549, opened 2021-11-01
feat(workspace-full): add ack-grep

ghuntley:
## Description

A better version of grep:

$ ack "string" *.md

## Release Notes

```release-note
- Add ack-grep to the base image
```

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -18,6 +18,7 @@ RUN yes | unminimize \
         locales \
         man-db \
         nano \
+        ack-grep \
         ripgrep \
         software-properties-common \
         sudo \


csweichel:
I don't think this should go into base. We want to keep base truly minimal. full would be a better home, if at all. How many folks know about this tool and would expect to find it by default?

ghuntley:
> I don't think this should go into base. We want to keep base truly minimal. full would be a better home, if at all. How many folks know about this tool and would expect to find it by default?

Moved to [full] from [base]. I use $ack day-to-day as often as you use $ dive :)

csweichel:
> > I don't think this should go into base. We want to keep base truly minimal. full would be a better home, if at all. How many folks know about this tool and would expect to find it by default?
> 
> Moved to [full] from [base]. I use `$ ack` day-to-day as often as you use `$ dive` :)

🙏 

Speaking of `dive`. I'd vote for removing it.

ghuntley:
> Speaking of `dive`. I'd vote for removing it.

**idea:** each Gitpod engineer gets to have _one_ (small) tool of choice inside workspace-full as a quality-of-life perk of being a Gitpodder? In implementation we would have a Gitpodder section at the end of the definition.

PR #637, opened 2022-01-21
Sync all images from GAR to dockerhub

princerachit:
Updated GAR -> Dockerhub using a dedicated step
    
Added all available images to the list of images that need to be synced.

### DIFF
--- a/.github/promote-images.yml
+++ b/.github/promote-images.yml
@@ -1,21 +1,21 @@
 # Sync all the image tags including latest
 europe-docker.pkg.dev/gitpod-artifacts/docker-dev:
   images:
-    workspace-base:
-    workspace-c:
-    workspace-clojure:
-    workspace-full:
-    workspace-full-vnc:
-    workspace-go:
-    workspace-node:
-    workspace-node-lts:
-    workspace-python:
-    workspace-ruby-2:
-    workspace-ruby-3:
-    workspace-rust:
-    workspace-dotnet-vnc:
-    workspace-postgresql:
-    workspace-mysql:
-    workspace-mongodb:
-    workspace-java-11:
-    workspace-java-17:
+    workspace-base: []
+    workspace-c: []
+    workspace-clojure: []
+    workspace-full: []
+    workspace-full-vnc: []
+    workspace-go: []
+    workspace-node: []
+    workspace-node-lts: []
+    workspace-python: []
+    workspace-ruby-2: []
+    workspace-ruby-3: []
+    workspace-rust: []
+    workspace-dotnet-vnc: []
+    workspace-postgresql: []
+    workspace-mysql: []
+    workspace-mongodb: []
+    workspace-java-11: []
+    workspace-java-17: []
--- a/.github/promote-images.yml
+++ b/.github/promote-images.yml
@@ -1,5 +1,21 @@
-# for now we intentionally put a hardcoded tag. Going forward we can make this empty to sync all tags. This is hardcoded so we don't endup updating the latest tag
+# Sync all the image tags including latest
 europe-docker.pkg.dev/gitpod-artifacts/docker-dev:
   images:
+    workspace-base:
+    workspace-c:
+    workspace-clojure:
     workspace-full:
-    - 2022-01-18-18-10-37
+    workspace-full-vnc:
+    workspace-go:
+    workspace-node:
+    workspace-node-lts:
+    workspace-python:
+    workspace-ruby-2:
+    workspace-ruby-3:
+    workspace-rust:
+    workspace-dotnet-vnc:
+    workspace-postgresql:
+    workspace-mysql:
+    workspace-mongodb:
+    workspace-java-11:
+    workspace-java-17:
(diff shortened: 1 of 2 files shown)


.github/workflows/image-promote.yml
```diff
@@ -1,11 +1,17 @@
-name: Sync workspace-full image tag from GAR to Dockerhub
+name: Sync images from GAR to Dockerhub
 
 on:
-  workflow_dispatch:
-    inputs:
-      tag:
-        description: 'worksapce-full image tag in GAR'
-        required: true
+  schedule:
```
kylos101, line 4: Let's talk about schedule, I'm not sure that we want to trigger every day just yet.

princerachit: Changed this to weekly. This would also mean that sync will take at least 1 week. Sure, let us discuss this offline

.github/promote-images.yml
```diff
@@ -1,5 +1,21 @@
...
+    workspace-python:
+    workspace-ruby-2:
+    workspace-ruby-3:
+    workspace-rust:
+    workspace-dotnet-vnc:
+    workspace-postgresql:
+    workspace-mysql:
+    workspace-mongodb:
+    workspace-java-11:
+    workspace-java-17:
```
kylos101, line 21: @princerachit I think each of these may need to be an array, rather than an empty value.

It may be helpful to run a test where you sync from GAR to a local docker registry. I'm just trying to avoid a scenario where we upload and then have to delete/clean-up.  WDYT?

From [the docs](https://github.com/containers/skopeo/blob/main/docs/skopeo-sync.1.md):

```
registry.example.com:
    images:
        busybox: []
        redis:
            - "1.0"
            - "2.0"
            - "sha256:0000000000000000000000000000000011111111111111111111111111111111"
    images-by-tag-regex:
        nginx: ^1\.13\.[12]-alpine-perl$
    credentials:
        username: john
        password: this is a secret
    tls-verify: true
    cert-dir: /home/john/certs
quay.io:
    tls-verify: false
    images:
        coreos/etcd:
            - latest
 ```

This will copy the following images:
- Repository registry.example.com/busybox: all images, as no tags are specified.
- Repository registry.example.com/redis: images tagged "1.0" and "2.0" along with image with digest "sha256:0000000000000000000000000000000011111111111111111111111111111111".
- Repository registry.example.com/nginx: images tagged "1.13.1-alpine-perl" and "1.13.2-alpine-perl".
- Repository quay.io/coreos/etcd: images tagged "latest".

Comment (princerachit) [reply .github/promote-images.yml:21]: Hey @kylos101 ,

> I'm just trying to avoid a scenario where we upload and then have to delete/clean-up

This PR will push ALL the tags that exist in GAR to DH. Do you want to restrict them? I don't mind publishing older tags. We officially announced the release a few days back and I expect customers to use tags starting from the day we made announcement.

If we want to restrict the sync then we have these options:
1. Delete the images that we do not want to publish manually from GAR
2. Modify the promote images yaml file on the fly to include only the latest and the current timestamp tag.
3. Only upload the new tagged image and the latest image

Personally I would not care for any of the above and rather sync everything. This also gives us advantage that if somehow we endup deleting an image from DH then the next sync will push it back

Comment (princerachit) [reply .github/promote-images.yml:21]: >@princerachit I think each of these may need to be an array, rather than an empty value.





Explanation:

`[]` is equivalent to not providing any tags.

e.g.
```yaml
localhost:8000:
  images:
    nginx:
```



will sync all the tags of nginx image:


```bash
➜  skopeo-work skopeo sync --registries-conf=insecure.conf --src-tls-verify=false --dest-tls-verify=false --src yaml --dest docker .sync.yaml localhost:6000
INFO[0000] Processing repo                               registry="localhost:8000" repo=nginx
INFO[0000] Querying registry for image tags              registry="localhost:8000" repo=nginx
INFO[0000] Getting tags                                  image="localhost:8000/nginx"
INFO[0000] Copying image ref 1/2                         from="docker://localhost:8000/nginx:latest" to="docker://localhost:6000/nginx:latest"
Getting image source signatures
Copying blob 5de8f8b958d2 done  
Copying blob 661b4150d3a3 done  
Copying blob 573040833908 done  
Copying blob 13b3c8cc8cde done  
Copying blob 8998bd30e6a1 done  
Copying blob 6fba654dd4ee done  
Copying config 2e7e2ec411 done  
Writing manifest to image destination
Storing signatures
INFO[0004] Copying image ref 2/2                         from="docker://localhost:8000/nginx:v1" to="docker://localhost:6000/nginx:v1"
Getting image source signatures
Copying blob 13b3c8cc8cde skipped: already exists  
Copying blob 6fba654dd4ee skipped: already exists  
Copying blob 573040833908 skipped: already exists  
Copying blob 8998bd30e6a1 skipped: already exists  
Copying blob 661b4150d3a3 skipped: already exists  
Copying blob 5de8f8b958d2 skipped: already exists  
Copying config 2e7e2ec411 done  
Writing manifest to image destination
Storing signatures
INFO[0004] Synced 2 images from 1 sources           
```

kylos101: I do not want to restrict.

I'm asking because the syntax we are currently using does not match skopeo's documentation for when you want to push everything.

kylos101 (approved):
/LGTM

:train2: :zap:

PR #660, opened 2022-02-03
Update composer and add tests

princerachit:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
```


sagor999 (approved):

kylos101 (approved):
@princerachit nice work! I wonder if the community would also like a php combination. :thinking:

Issue #675, opened 2022-02-09
gitpod/workspace-mysql no longer installs apache2


apolopena:
### Bug description

Any `.gitpod.Dockerfile` using:
```yml
FROM gitpod/workspace-mysql
```
or 
```yml
FROM gitpod/workspace-mysql:latest
```
no longer has apache2 installed.

This breaks alot of things. 

All docker image build steps will break if an assumed directory such as /etc/apache2/ is used to lets say copy a `apache2.conf ` file over or something like that which is a typical thing to do.

Note: `workspace-full` is unaffected.

### Steps to reproduce

Create a workspace from the url below which is a copy of https://github.com/gitpod-io/apache-example
with one single change: https://github.com/apolopena/gitpod-bug-8113/commit/b6148ea255e817f580b1c282f71d7db08ae809e3

https://gitpod.io/#/https://github.com/apolopena/gitpod-bug-8113

Terminal output: `bash: apachectl: command not found`
Running: `apache2 -v` gives the error:
```log
bash: apache2: command not found
```
Apache2 is not installed.

### Workspace affected

Any workspace that uses gitpod/workspace-mysql

### Expected behavior

Any image built on `workspace-full` such as `workspace-mysql` should have apache2. Removing this is not backwards compatible.

### Example repository

https://github.com/apolopena/gitpod-bug-8113

### Anything else?

In order to maintain vital backwards compatibility in the gitpod workspace images, things like this should not change.

apolopena:
@jimmybrancaccio
Maybe you can shed some light on this? Some workaround that doesn't involved implementing my own mysql integration on top of workspace-full?
https://github.com/apolopena/gitpod-laravel-starter/issues/200#issuecomment-1034165012

jimmybrancaccio:
@apolopena

> Unfortunately legacy-dazzle-v1 is buggy with ca certificates and should not be used.

I wouldn't say it's buggy, as there's no bug. However you do need to manually update the packages as I noted in #599.

> So now it seems that I should probably maintain my own image and work Gitpods changes in as I see fit.

You can definitely do this as well, the joys of open source!

I will probably end up having to build my own images as well since I build with Laravel and need MySQL and Apache too.

princerachit:
Thanks for pointing this out. I am working internally with the team on this and would update you soon.

davekiss:
[...]

PR #677, opened 2022-02-10
Use full image for combinations derived from legacy images

princerachit:
## Description

Change the base images to workspace-full for the images that had full in the legacy image.

## Related Issue(s)

Fixes https://github.com/gitpod-io/workspace-images/issues/675

## How to test

NA

## Release Notes

```release-note
Use full image as base for dotnet-vnc, mysql, postgreql and mongodb
```

### DIFF
--- a/dazzle.yaml
+++ b/dazzle.yaml
@@ -72,23 +72,23 @@ combiner:
         - lang-rust
     - name: dotnet-vnc
       ref:
-      - base
+      - full
       chunks:
         - tool-vnc
         - tool-dotnet
     - name: postgresql
       ref:
-      - base
+      - full
       chunks:
         - tool-postgresql
     - name: mysql
       ref:
-      - base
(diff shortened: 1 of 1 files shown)


dazzle.yaml
```diff
@@ -72,23 +72,23 @@ combiner:
         - lang-rust
     - name: dotnet-vnc
       ref:
-      - base
+      - full
```
Furisto, line 75: Why does dotnet need 'full'?

kylos101: Good question. It used to have full as its base image. In other words, it's a breaking change to users, to change the base image for dotnet from full to base, because they have dependencies on things that were part of full even while using the prior dotnet image.

For now, we should fix the regression, and later either plan a separate effort to make slimmer images (or build them dynamically from chunks altogether).

PR #685, opened 2022-02-14
feat(nix): bump version to 2.6.0

emilytrau:
## Description
Upgrades the version of nix to the latest release (2.6.0). This includes the major 2.4 update, which introduced support for features such as flakes.

## Related Issue(s)

N/A

## How to test

Open a workspace with this image
```shell
gitpod ~ $ nix --version
nix (Nix) 2.6.0
gitpod ~ $ nix run "nixpkgs#hello"
Hello, world!
gitpod ~ $ 
```

## Release Notes

```release-note
- Bump nix to 2.6.0
```

### DIFF
--- a/chunks/tool-nix/Dockerfile
+++ b/chunks/tool-nix/Dockerfile
@@ -2,6 +2,7 @@ ARG base
 FROM ${base}
 
 ARG NIX_VERSION
+ARG NIX_CONFIG=""
 
 ENV NIX_VERSION=${NIX_VERSION}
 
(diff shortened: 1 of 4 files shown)


princerachit:
Thanks for the PR. Can you add an additional variant instead of modifying the existing one? as you mentioned:
>This includes the major 2.4 update

This will make it easy for us to change if we want to publish older the nix version with `full` image

emilytrau:
@princerachit no worries. added that back in :)

chunks/tool-nix/Dockerfile
```diff
@@ -2,6 +2,7 @@ ARG base
 FROM ${base}
 
 ARG NIX_VERSION
+ARG NIX_CONFIG=""
```
princerachit, line 5: You are setting this to empty value and then later using it to write to a file. However, this value is also being supplied from  chunk file. How would this behave?

tests/tool-nix.yaml
```diff
@@ -8,4 +8,5 @@
   entrypoint: [bash, -i, -c]
   assert:
   - status == 0
-  - stdout.indexOf("2.3.14") != -1
+  - stdout.indexOf("2.3.14")  != -1 ||
```
princerachit, line 11: check the spacing in the two lines

chunks/tool-nix/Dockerfile
emilytrau: This is to let variants specify optional additional nix config options. Some features in new versions need to be enabled with an experimental flag, but this can cause older versions of nix to show a warning if those flags are defined and unknown.

princerachit (approved):

PR #686, opened 2022-02-14
Adding ability to specify a custom user

### LINKED ISSUE #695: Ability to use a custom  user instead of gitpod
I would like to use workspace-image with a custom user as of now the image comes with gitpod user.

mwaaas:
## Description
Using gitpod workspace image with a custom username 
Closes #695 
## Related Issue(s)

Fixes #

## How to test
- Build docker with user build args 
 docker build   --build-arg USER=mwaaas    --tag=test -f base/Dockerfile ./base/
 
 then docker run test $(whoami) should print mwaaas

repeat the same process without specifying user it should print gitpod

## Release Notes

```release-note
```
- Possible to specify user on build

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -4,6 +4,7 @@ COPY install-packages upgrade-packages /usr/bin/
 
 ### base ###
 ARG DEBIAN_FRONTEND=noninteractive
+ARG USER=gitpod
 
 RUN yes | unminimize \
     && install-packages \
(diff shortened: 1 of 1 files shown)


axonasif:
Hey @mwaaas, thanks for raising this PR.
There is a problem though, we have references on other places of `USER gitpod` and just changing here on `base` image won't cut it.
https://github.com/gitpod-io/workspace-images/blob/7d640a13abc662a459f885f7a432e8d9337d9a4e/chunks/lang-c/Dockerfile#L21

kylos101:
https://github.com/gitpod-io/workspace-images/issues/695#issuecomment-1058758406

mwaaas:
Hey @axonasif, 
So the change that I had was adding USER as argument but if you don't pass it during build it was picking gitpod as the default user. 

That won't break the other places where USER has not been added.

PR #712, opened 2022-02-20
refactor(chunks/tool-nginx): Add major php versions and extensions

### LINKED ISSUE #638: ppa:ondrej/php should not be the default PPA and is not used as recommended by the author for apache2 and nginx
### Bug description

The latest [PHP update from 7.4 to 8.0](https://github.com/gitpod-io/workspace-images/commit/0132229c6b1c552009b4432f7612163f84dc5e96#diff-e9ad2eb43016881f65b443658a8d94f33a4f3750743e33326c529f899be708d3R28) in [gitpod/workspace-images/full](https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile) switches the PHP repository to the untrusted (but trustable ;) [`ppa:ondrej/php`](https://launchpad.net/~ondrej/+archive/ubuntu/php). 

1. If a user wants the added functionality (such as running different versions of PHP side by side) of this PPA they should probably install it themselves and let the default installation be the official Ubuntu package at http://archive.ubuntu.com/ubuntu, as this installation will probably be the more common setup found on production machines.
2. The [`ppa:ondrej/php`](https://launchpad.net/~ondrej/+archive/ubuntu/php) page states that there are caveats when running this version of the PHP packages with apache2 and nginx. 
```
CAVEATS:
1. If you are using php-gearman, you need to add ppa:ondrej/pkg-gearman
2. If you are using apache2, you are advised to add ppa:ondrej/apache2
3. If you are using nginx, you are advised to add ppa:ondrej/nginx-mainline
   or ppa:ondrej/nginx
```
3.  [gitpod/workspace-images/full](https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile) is not currently following the recommended workarounds for running apache2 and nginx which is to use the ondrej PPA versions of apache2 and ngnix.

### Steps to reproduce

1. Open any modern default gitpod temaplate such as https://github.com/gitpod-io/template-typescript-react
2. Run the following command from terminal and make note of the PPA urls.
```bash
apt-cache policy apache2 && apt-cache policy nginx && apt-cache policy php8.0
```
Output should looks something like this:
image

### Workspace affected

any workspace that uses or derives from the image https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile

### Expected behavior

[`ppa:ondrej/php`](https://launchpad.net/~ondrej/+archive/ubuntu/php) should not be used as the default PHP installation in [gitpod/workspace-images/full](https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile)

If the maintainers of [gitpod/workspace-images/full](https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile) insist on keeping [`ppa:ondrej/php`](https://launchpad.net/~ondrej/+archive/ubuntu/php) as the default installtion of PHP in [gitpod/workspace-images/full](https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile) then the [ppa:ondrej/php caveats](https://launchpad.net/~ondrej/+archive/ubuntu/php) should be followed and apache2 and nginx should use the ondrej PPA's for those packages (which is currently not the case).

### Example repository

_No response_

### Anything else?

Using `ppa:ondrej/php` by default but not as recommended by the author on the [launchpad page](https://launchpad.net/~ondrej/+archive/ubuntu/php) does not seems like the way to go but please let me know your thoughts on this.
Thank you.

### LINKED ISSUE #714: workspace-mysql  has regressed to using PHP 7.4
### Bug description

`workspace-mysql` was bumped up to PHP 8.0 last month in the underlying `workspace-full` image and has recently regressed to using PHP 7.4

### Steps to reproduce

1. Ensure the docker image layer cache has been busted
2. Create any workspace from a repo that uses `workspace-mysql` in its `.gitpod.Dockerfile`
3. Run `php -v` in the terminal once the workspace has built and IDE has loaded.

### Expected behavior

PHP version should be 8.0

### Example repository

Any repository that  has:
```.dockerFile
FROM gitpod/workspace-mysql
```
in its `.gitpod.Dockerfile`

### Anything else?

It is expected that the PHP version will bump up from time to time but the PHP version should not ever bump down.

axonasif:
## Description

Installs all major PHP versions (including 8.x as previously).

## Related Issue(s)

Fixes #714
See relevant [discussion on discord.](https://canary.discord.com/channels/816244985187008514/944878139652313169/945010959938904175)

## Release Notes

```release-note
```


apolopena:
This does resolve https://github.com/gitpod-io/workspace-images/issues/638 but perhaps resolving this the other way around (remove `ppa:ondrej`) would be more prudent and would reduce potential regressions caused by changing the PPA for Apache _and_ nginx. It is my understanding that the primary reason for using `ppa:ondrej` is for installing multiple versions of PHP side by side and that would probably be best left up to the user since it is a bit of an edge case.  Less change in the image to maintain utmost backward compatibility seems like a good thing here.

Also it might be a good idea to be explicit here with the PHP version number such as PHP8.0. At a glance it is not clear what version of PHP is being installed. Could this cause #714?

Lastly how about decoupling Apache from nginx and possibly PHP as well? This would make for better testing and allow for slim worskpaces like `workspace-nginx` and `workspace-apache` in the future. If we want to keep these 'servers' coupled then perhaps renaming `tool-nginx` to `tool-servers` for the sake of clarity would be a good idea.

chunks/tool-nginx/Dockerfile
```diff
@@ -6,14 +6,16 @@ USER root
 # Dazzle does not rebuild a layer until one of its lines are changed. Increase this counter to rebuild this layer.
 ENV TRIGGER_REBUILD=1
 
-RUN install-packages \
+RUN for _ppa in 'ppa:ondrej/php' 'ppa:ondrej/apache2' 'ppa:ondrej/nginx-mainline'; do add-apt-repository -y "$_ppa"; done \
+    && install-packages \
```
princerachit, line 10: I am bit confused about the PR. The associated issue says not to use `ppa:ondrej/php` but that is still being added here. And +1 to the suggestion of specifying the version of the PHP.

princerachit (commented):
Please clarify

chunks/tool-nginx/Dockerfile
axonasif: Perhaps we create multi-version docker images?

kylos101: @axonasif yes, Dazzle variants make sense for PHP 7.4 and 8.0, which'll result in separate images being published to Docker Hub. 

Also, `ppa:ondrej/php` should be removed, I think. I am going to flip this PR back to draft for now. 

Is this something you would still like to own? If yes, what do you think of the new scripts to build and test chunks and combinations?

axonasif: Hey @princerachit and @kylos101 👋 
I haven't been able to find any php8.0(or X) package in the official ubuntu and/or debian repositories.
It seems that for both cases [Ondřej Surý](https://askubuntu.com/questions/1333214/can-i-trust-a-repo-i-am-unfamiliar-with-to-provide-me-a-secure-php-8-0-package) comes in.

Perhaps @pawlean and I can try creating a poll on our discord server to see if we get any input regarding this, thanks for the idea 🙏

axonasif: > Lastly, is this something you would still like to own? 🤔 If yes, what do you think of the new scripts to build and test chunks and combinations? 🙏

Yeah, happy to collab on this PR 🙏 
I think that it's much better now since we can save our time, however I also think that it would be nicer if we had one `dazzle` wrapper script for all tasks with subcommands baked in, instead of multiple `dazzle-*.sh` files 😄

kylos101: The intent of getting the scripts out there was to ship skateboards. If we find having many files is proves problematic, we can certainly iterate to condense in the future.

kylos101:
Hi @axonasif :wave: , it seems like this change will fix #714, and we should (in hindsight) close #638 (given what we've learned in this PR about Focal and php8).

I am going to update this PR accordingly (change the description and linkage from 638 to 714). Cool?

Also, if this is ready for review now, feel free to click the button and we'll take it from there. :+1:

kylos101:
@axonasif the only thing I think we're missing, are [tests](https://github.com/gitpod-io/workspace-images/blob/axon/tool-nginx-patch/tests/tool-nginx.yaml) to assert the version of php in the `tool-nginx` chunk. May I ask you to copy [these tests](https://github.com/gitpod-io/workspace-images/blob/axon/tool-nginx-patch/tests/lang-php.yaml), and update the version check from 7 to 8 in the tests for the`tool-nginx` tests?

Separate from this issue and PR, if the community needs variant images for PHP (like 7 versus 8), we can handle in a separate issue/PR.

apolopena:
How does one tell at a glance from this file what version of PHP is being used?
Why not use the explicit PHP version number?

kylos101:
> How does one tell at a glance from this file what version of PHP is being used? Why not use the explicit PHP version number?

Hi @apolopena :wave: , excellent question, thank you for reaching out, and also for your help with this issue. The intent of this PR is to update the `tool-nginx` chunk (which is used in many images, like `workspace-full`, which was upgraded from PHP7 to PHP8 last year, but regressed). 

It sounds like you'd prefer a couple images for php (in general). May I ask you to open a separate issue to describe what it is that you're looking for? For example, it sounds like you're asking for `workspace-php-7` and `workspace-php-8`. I ask for a separate issue because we'll want to discuss boundaries for those images. For example, the base will be `workspace-base` (which includes Docker and Tailscale), and we'll install PHP, but, not a database or other tools. Users that wish for a database are recommended to use a custom docker file and layer on their own customizations.

apolopena:
> > How does one tell at a glance from this file what version of PHP is being used? Why not use the explicit PHP version number?
> 
> Hi @apolopena 👋 , excellent question, thank you for reaching out, and also for your help with this issue. The intent of this PR is to update the `tool-nginx` chunk (which is used in many images, like `workspace-full`, which was upgraded from PHP7 to PHP8 last year, but regressed).
> 
> It sounds like you'd prefer a couple images for php (in general). May I ask you to open a separate issue to describe what it is that you're looking for? For example, it sounds like you're asking for `workspace-php-7` and `workspace-php-8`. I ask for a separate issue because we'll want to discuss boundaries for those images. For example, the base will be `workspace-base` (which includes Docker and Tailscale), and we'll install PHP, but, not a database or other tools. Users that wish for a database are recommended to use a custom docker file and layer on their own customizations.

Actually I solved the issue caused by Gitpod bumping up (and down) the version of PHP from time to time which makes sense, well the bumping up part anyway :smile:. I solved the issue in my projects by dynamically installing PHP the `ppa` and xdebug from a config file if the PHP version or `ppa`  in the Gitpod docker image is not what the user wants.

So for me individual images with different versions of PHP and a database I have to setup myself is not helpful. I just want to be able to know the version of PHP that Gitpod uses by default and currently it is set implicitly.

Lastly for the sake of backward compatibility please leave `workspace-mysql` using `workspace-full` and `workspace full` using `tool-nginx` along with `node`, `go` and the whole crew of tools many of us depend on and would have to rewrite _and test _ sunstantial amounts of code in order to aleviate breaking changes and regressions. Top level legacy images like `workspace-mysql` should be left alone as much as possible for the reasons I have stated and probably a few more I haven't thought of.

I know it makes sense for Gitpod to offer out-of-the-box docker images in small pieces that work togther rather than monolithic ones and I fully agree with this. I am happy to suggest small images I would find useful that I might think of in a separate issue.

Thank you.

kylos101 (approved):
Hey @axonasif thanks a lot for fixing #714!

Hey @apolopena we really appreciate your thoughtful and detailed feedback. 

[@axonasif added this test to assert the version of PHP in the tool-nginx chunk](https://github.com/gitpod-io/workspace-images/pull/712/commits/5edcf56fa89d583f4318b8648d5d10bf914e5dbb#diff-32c0277d131125e335703271768cfde1528ca4113e162ea85b3f709478f0b294R11). 

The test (1) is explicit about the PHP version and (2) should help us avoid regressions in the future.

FYI, if anyone needs PHP 7, they'll have to pursue an alternative. For example, use an older tagged image support PHP7, or use a workaround as you described.

axonasif:
Hey @kylos101, thanks for the fast approval 🙏 
> FYI, if anyone needs PHP 7, they'll have to pursue an alternative. For example, use an older tagged image support PHP7, or use a workaround as you described.

Actually, when we install PHP with Ondrej PPA, we get all the versions installed side by side 🚀
Screenshot 2022-04-19 at 10 54 53 PM

Issue #713, opened 2022-02-21
Use Google Artifact Registry to Cache images and chunks


princerachit:
## Current Status
At the moment we are using Github Action's Caching in order to store the generated chunks and images for a branch when a [PR is opened](https://github.com/gitpod-io/workspace-images/blob/b88ed0d2b6a4a0adbf4fdaa9394440327b33d53b/.github/workflows/pull-request.yml#L38-L50).

We also use this caching for the [master branch](https://github.com/gitpod-io/workspace-images/blob/b88ed0d2b6a4a0adbf4fdaa9394440327b33d53b/.github/workflows/push-main.yml#L66-L78).

## Caveats
1. The maximum size of a cache for one repository is [only 10Gi](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy). A few images can take up 10Gi so the cache is never available

## Proposed solution
Instead of using GH Actions Cache, we can use Google Artifact registry to store the images and layers that are produced during build. GAR does not have any restriction on the size of images.

Since we would want to have dedicated images for each PR we can use images with branch suffixes/tags. e.g. `repo-name/workspace-images-<branch>` or `repo-name/workspace-images:<branch>`

For master branch we can do the same except the use of branch name.

We can create another GH action that deletes the temporary images which were generated due to a Pull request.

## Tasks
- [x] Update push-main.yml workflow to make use of GAR as image registry
- [x] #728
- [x] Update pull-request.yml workflow to make use of GAR as image registry
- [ ] <del>Create a merge-pr.yml workflow which deletes the images generated from a PR with branch name <branch-name> </del>
- [x] Create a schedule-pr-reg-clueanup workflow which deletes all the image tag in docker-dev-pr registry which are older than X days
- [ ] <del>#760</del> removed from this issue

princerachit:
Blocked on https://github.com/gitpod-io/workspace-images/issues/741

PR #719, opened 2022-02-21
Use GAR to build images for pull-request.yml workflow

princerachit:
## Description

Use GAR to build images instead of local registry.
This would enable faster builds as GAR would house the previously built chunks.
Modify dazzle.yaml file to rename the combination so that two PRs do not race to write to the same tag.
Use PR number suffix to distinguish builds.

## Related Issue(s)

Fixes partially https://github.com/gitpod-io/workspace-images/issues/713

## How to test

See the pull-request.yml action for this PR. It is already using GAR.

## Release Notes

```release-note
None
```

### DIFF
--- a/.github/update-dazzle-for-pr.py
+++ b/.github/update-dazzle-for-pr.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python3
+import sys
+import shutil
+from ruamel.yaml import YAML
+
+
+n = len(sys.argv)
+if n != 2:
+    raise Exception(
+        "mandatory argument <suffix> missing, use update-dazzle-for-pr.py <suffix>")
+
+suffix = sys.argv[1]
+
+if len(suffix) == 0:
+    raise Exception("mandatory argument <suffix> should not be empty")
+
+# take backup
+shutil.copy2("dazzle.yaml", "dazzle.yaml.orig")
+
+yaml = YAML(typ="safe")
+
+fp = open("dazzle.yaml.orig")
+data = yaml.load(fp)
+fp.close()
+
+# add suffix '-<suffix>' to all the combo names and the references
+for combo in data["combiner"]["combinations"]:
+    combo["name"] += "-" + suffix
+    if "ref" in combo:
+        for i in range(len(combo["ref"])):
+            combo["ref"][i] += "-" + suffix
+
+final_fp = open("dazzle.yaml", "w")
+yaml.dump(data, final_fp)
+final_fp.close()
--- a/.github/update-dazzle-for-pr.py
+++ b/.github/update-dazzle-for-pr.py
@@ -19,9 +19,9 @@ shutil.copy2("dazzle.yaml", "dazzle.yaml.orig")
 
 yaml = YAML(typ="safe")
 
-fp = open("dazzle.yaml.orig")
-data = yaml.load(fp)
-fp.close()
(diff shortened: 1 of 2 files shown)


.github/update-dazzle-for-pr.py
```diff
@@ -0,0 +1,35 @@
...
+    raise Exception(
+        "mandatory argument <suffix> missing, use update-dazzle-for-pr.py <suffix>")
+
+suffix = sys.argv[1]
+
+if len(suffix) == 0:
+    raise Exception("mandatory argument <suffix> should not be empty")
+
+# take backup
+shutil.copy2("dazzle.yaml", "dazzle.yaml.orig")
```
utam0k, line 18: 👍

.github/update-dazzle-for-pr.py
```diff
@@ -0,0 +1,35 @@
...
+
+yaml = YAML(typ="safe")
+
+fp = open("dazzle.yaml.orig")
+data = yaml.load(fp)
+fp.close()
```
utam0k, line 24: How about this?
```suggestion
with open("dazzle.yaml.orig") as fp:
    data = yaml.load(fp)
```

princerachit:
>What happens if a malicious user creates a large number of PRs?

Nothing, we manually approve PR runs. 
image
And the settings are [here](https://github.com/gitpod-io/workspace-images/settings/actions).

utam0k:
> >What happens if a malicious user creates a large number of PRs?
> 
> Nothing, we manually approve PR runs. 
> image
> And the settings are [here](https://github.com/gitpod-io/workspace-images/settings/actions).

Nice!

utam0k (approved):

princerachit:
Merge this only when https://github.com/gitpod-io/workspace-images/pull/729 is merged

.github/workflows/pull-request.yml
```diff
@@ -33,30 +41,34 @@ jobs:
...
           sudo chmod +777 /run/buildkit/buildkitd.sock
 
-      # A hack as GH action does not allow you to force override cache storing if there was a cache hit
-      # https://github.com/actions/cache/issues/628#issuecomment-986118455
-      - name: 🗄️ Force Save Registry Cache Per Sha
-        uses: actions/cache@v2
+      - name: 🤓 Update dazzle.yaml
+        run: |
+          pip3 install ruamel.yaml
+          python3 .github/update-dazzle-for-pr.py ${{ github.event.pull_request.number }}
```
kylos101, line 47: We often use `yq` to make changes to YAML, out of curiosity,m why not use it, instead of Python, to add the suffix?

Full disclosure, I like Python, I'm just curious.

kylos101: [Example using jq and yq](https://github.com/gitpod-io/gitpod/blob/main/.werft/jobs/build/deploy-to-preview-environment.ts#L351-L355) (granted, this yq version is older)

princerachit: Hey @kylos101 , I did explore the option of using `yq` but the query was unnecessarily complex to understand and write. I figured a python script was much simpler to write and understand.

princerachit:
I am merging this PR to unblock other PR work. @kylos101 if you still think we should switch to using `yq` then we can discuss it on a separate forum

PR #722, opened 2022-02-22
Delete older image tags and layers for docker-dev-pr/workspace-base-images-pr image

princerachit:
## Description

Every PR results in creation of images and layers. These are stored in the GAR. We don't need layers created from old PRs which are merged. Since there is no way to identify the sha of a layer based on the PR, we delete all the layers and tags in the GAR which are older than 4 days. In some cases when a PR is open longer than 4 days a manual retrigger of build or a new push to the PR will result in rebuilding of layers. This is not so bad considering the amount of storage the stale layers will consume.

## Related Issue(s)

Partially Fixes https://github.com/gitpod-io/workspace-images/issues/713

## How to test

I tested the changes by [triggering them on a pull_request](https://github.com/gitpod-io/workspace-images/runs/5350633387?check_suite_focus=true) event and it succeeded.
## Release Notes

```release-note
None
```

### DIFF
--- a/.github/gargc.sh
+++ b/.github/gargc.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# Inspired from https://gist.github.com/ahmetb/7ce6d741bd5baa194a3fac6b1fec8bb7
+
+IFS=$'\n\t'
+set -eou pipefail
+
+if [[ "$#" -ne 2 || "${1}" == '-h' || "${1}" == '--help' ]]; then
+	cat >&2 <<"EOF"
+gargc.sh cleans up tagged or untagged images pushed before specified date
+for a given image
+USAGE:
+  gargc.sh REPOSITORY DATE
+EXAMPLE
+  gargc.sh  europe-docker.pkg.dev/gitpod-artifacts/docker-dev-pr/workspace-base-images-pr 2017-04-01
+  would clean up all images tags pushed before 2017-04-01 for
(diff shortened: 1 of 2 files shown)


.github/gargc.sh
```diff
@@ -0,0 +1,39 @@
...
+	exit 1
+elif [[ "${#2}" -ne 10 ]]; then
+	echo "wrong DATE format; use YYYY-MM-DD." >&2
+	exit 1
+fi
+
+main() {
+	local C=0
+	IMAGE="${1}"
+	DATE="${2}"
```
utam0k, line 28: Maybe it would be better if we can specify `>= 2017-04-01` or `> 20217-04-01`. The reason for this is to make it clearer whether it is included or not. If it could be done easily, it would be wonderful.

princerachit: I have included the information that the deletion is for all layers pushed **before** a certain date. So, the date is not included. I also updated the description to reflect this.

.github/workflows/pr-registry-cleanup.yml
```diff
@@ -0,0 +1,44 @@
+name: Delete PR registry images
+on:
+  schedule:
+  # At 04:07 everyday. We choose such time as github recommends not choosing hourly timings
+  # like 04:00 to avoid schedule job congestions
```
utam0k, line 5: 👍

utam0k (approved):

Issue #728, opened 2022-02-25
Separate workflow which pushes to Docker Hub


kylos101:
## Is your feature request related to a problem? Please describe

We currently push to Docker Hub each time we merge to the default branch.

## Describe the behaviour you'd like

- We do not want to push to Docker Hub with each merge to the default branch.
- We want to trigger a push of content to Docker Hub separately, perhaps weekly.
- In this way, there will be less versions of `workspace-full` floating around out there, and less images that much be cached on nodes.

## Additional context

#713

kylos101:
Hey @princerachit may I ask you to work this issue, prior to the other items in #713, this way it gets merged to main beforehand? That will give us the ability to merge to main without always pushing to Docker Hub.

CC: @utam0k for awareness

princerachit:
>In this way, there will be less versions of workspace-full floating around out there, and less images that much be cached on nodes.

At this moment we sync all image tags from GAR to DH. Even if we push weekly we would be shipping multiple versions of workspace-full. I guess you mean that there should be less changes in the `latest` tag of workspace full.

PR #729, opened 2022-02-25
Sync images tags to Dockerhub using manual Action

princerachit:
### Original Behaviour
The `push-main.yml` workflow would build new images whenver there is a merge to the master branch.
It will rename and copy the images along with datetimestamp to GAR.
It will also retag the new image tag with `latest` tag.
It will push all the tags of images in GAR to Dockerhub via skopeo sync.

### New behaviour
Do not push new image tags from GAR to Docker hub when there are merges to master.
Create a new workflow `dockerhub-release.yml` which can be only triggered manually.
This workflow will copy all the tags including the `latest` tag of workspace images to Docker Hub from GAR.

## Other Changes
Additionally -  I removed the `registries.conf` file which is no more applicable as we switched to GAR.

## Related Issue(s)

Fix https://github.com/gitpod-io/workspace-images/issues/728

## How to test

NA. Can be only tested after merge.

## Release Notes

```release-note
None
```

### DIFF
--- a/.github/workflows/dockerhub-release.yml
+++ b/.github/workflows/dockerhub-release.yml
@@ -1,0 +1,73 @@
+name: Publish latest images to Docker Hub
+
+on:
+  workflow_dispatch:
+
+jobs:
+  # Sync the 'latest' tag from GAR to Docker Hub
+  #
+  # From your GitHub repo clock Settings. In the left menu, click Environments.
+  # Click New environment, set the name production, and click Configure environment.
+  # Check the "Required reviewers" box and enter at least one user or team name.
+  promote-latest:
+    runs-on: ubuntu-latest
+    environment: "production"
+    permissions:
+      contents: "read"
+      id-token: "write"
+    env:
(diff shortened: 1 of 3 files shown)


.github/workflows/dockerhub-release.yml
```diff
@@ -0,0 +1,73 @@
...
+        id: "auth"
+        uses: "google-github-actions/auth@v0.4.3"
+        with:
+          token_format: "access_token"
+          workload_identity_provider: ${{env.WORKLOAD_IDENTITY_POOL_ID}}
+          service_account: ${{env.IAM_SERVICE_ACCOUNT}}
+
+      - name: ✍🏽 Login to GAR using skopeo
+        run: |
+          sudo skopeo login -u oauth2accesstoken --password=${{ steps.auth.outputs.access_token }} https://${{env.GAR_IMAGE_REGISTRY}}
```
sagor999, line 59: this will not leak the password in some log or output field? (just double checking)

sagor999: I guess not, since I see it was used like that in push-main.yml as well.

sagor999 (approved):

PR #730, opened 2022-02-25
Prs/release datetimestamp

princerachit:
### Original Behaviour
The `push-main.yml` workflow would build new images whenver there is a merge to the master branch.
It will rename and copy the images along with datetimestamp to GAR.
It will also retag the new image tag with `latest` tag.
It will push all the tags of images in GAR to Dockerhub via skopeo sync.

### New behaviour
Sync only date time stamp tagged images from GAR to dockerhub using skopeo.
Create a new workflow `dockerhub-release.yml` which can be only triggered manually.
This workflow will copy only the `latest` tag of workspace images to Docker Hub from GAR.

So, everytime there is a merge to the master branch new set of images will be release to public on Docker Hub, however, it will not update the latest tag.
You can see [skopeo doc](https://github.com/containers/skopeo/blob/main/docs/skopeo-sync.1.md) to learn more on the yaml changes.

## Other Changes
Additionally -  I removed the `registries.conf` file which is no more applicable as we switched to GAR.

## Related Issue(s)

Addresses https://github.com/gitpod-io/workspace-images/issues/728

## How to test

NA. Can be only tested after merge.

However, I tested if the regex matching works locally by running skopeo locally.

## Release Notes

```release-note
None
```

### DIFF
--- a/.github/promote-images.yml
+++ b/.github/promote-images.yml
@@ -1,21 +1,23 @@
-# Sync all the image tags including latest
+# Sync all the iamges that are tagged using date time stamp
+# format +%Y-%m-%d-%H-%M-%S e.g. 2022-02-22-10-10-54
+# Use simple regex to avoid confusion. Below will work for 20xx years.
 europe-docker.pkg.dev/gitpod-artifacts/docker-dev:
-  images:
-    workspace-base: []
-    workspace-c: []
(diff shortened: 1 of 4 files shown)


princerachit:
closing in favour of https://github.com/gitpod-io/workspace-images/issues/728

PR #735, opened 2022-02-27
Update accesstoken lifetime to 5 hours

princerachit:
## Description

Update the default timeout of google access token to 18000s or 5 hours to avoid access token expiration for a longer running action.

## Related Issue(s)

Partially Fixes https://github.com/gitpod-io/workspace-images/issues/713

## How to test

The PR build should be green. Nothing to be tested manually.

## Release Notes

```release-note
None
```

### DIFF
--- a/.github/workflows/dockerhub-release.yml
+++ b/.github/workflows/dockerhub-release.yml
@@ -51,6 +51,7 @@ jobs:
         uses: google-github-actions/auth@v0.6.0
         with:
           token_format: "access_token"
+          access_token_lifetime: '18000s'
           workload_identity_provider: ${{env.WORKLOAD_IDENTITY_POOL_ID}}
           service_account: ${{env.IAM_SERVICE_ACCOUNT}}
 
--- a/.github/workflows/image-promote.yml
+++ b/.github/workflows/image-promote.yml
@@ -47,6 +47,7 @@ jobs:
         uses: google-github-actions/auth@v0.6.0
         with:
           token_format: "access_token"
+          access_token_lifetime: '18000s'
           workload_identity_provider: ${{env.WORKLOAD_IDENTITY_POOL_ID}}
           service_account: ${{env.IAM_SERVICE_ACCOUNT}}
 
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -55,7 +55,7 @@ jobs:
 
       - name: 🔐 Authenticate to Google Cloud
         id: "auth"
-        uses: google-github-actions/auth@v0.6.0
+        uses: google-github-actions/auth@v0
         with:
           token_format: "access_token"
           access_token_lifetime: '18000s'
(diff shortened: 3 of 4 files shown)


kylos101:
Hey @princerachit nice find. Can you share some background on which action(s) and steps require such a long lived token? 

I ask because five hours seems like a long time, and we might have an opportunity to improve whatever is going slowly.

princerachit:
> Hey @princerachit nice find. Can you share some background on which action(s) and steps require such a long lived token?
> 
> I ask because five hours seems like a long time, and we might have an opportunity to improve whatever is going slowly.
> 
> edit: [is it because of this failure](https://github.com/gitpod-io/workspace-images/actions/runs/1906312853)?

e.g. failure https://github.com/gitpod-io/workspace-images/actions/runs/1880495134, For the first time builds(when no chunk is cached) I believe it can take 2hours+ during dazzle build. I don't want to tweak this value if the build fails the next time because it took 2 hours and 1 minute.
Post `dazzle build` we also do `dazzle combine` which again needs access token to talk to GAR. So the token should be valid for the worst case of these combined times.

kylos101:
@princerachit will every PR take 2+ hours to `dazzle build` the first time, say for example if they just changed a single test file? Or is that just worst case scenario depending on how many chunk files are changed?

princerachit:
> @princerachit will every PR take 2+ hours to dazzle build the first time, say for example if they just changed a single test file? Or is that just worst case scenario depending on how many chunk files are changed?

No, every PR won't take that much time. It is the worst case scenario. This worst case scenario can occur as [often as 4 days](https://github.com/gitpod-io/workspace-images/pull/722), That's is the duration after which we clear out image layers/chunks from GAR using automated jobs.
So, If I raise a PR today which builds all the chunks and caches them in GAR then all subsequent PRs will reuse the cache chunks depending on the file changes. After 4 days we will clear out these cached chunks. If a new PR is opened, it will try building these chunks again and can take 2hours+. Any subsequent PR for the next 4 days will reuse these new cached chunks.

We cannot determine which chunk layers/tags should be long lived so we clear anything older than 4 days.

aledbf:
> We cannot determine which chunk layers/tags should be long lived so we clear anything older than 4 days.

Why 4 days?

princerachit:
> > We cannot determine which chunk layers/tags should be long lived so we clear anything older than 4 days.
> 
> Why 4 days?

No specific reason. The number is arbitrary.

aledbf:
> No specific reason. The number is arbitrary.

Ok. Can we switch that to something like 30 days?

princerachit:
>Ok. Can we switch that to something like 30 days?

I feel like 30 days is too much. Maybe not very evident but it somehow indicates that a PR can remain open for 30 days.

princerachit:
I am moving this to draft state as the GH action is complaining about missing permission which is already assigned to the service accounts. I have opened an[ issue in the auth action](https://github.com/google-github-actions/auth/issues/145) repo

kylos101:
Hey @princerachit , I think for now we should rely on the local docker registry in the PR.

Why? Once [this is done](https://github.com/gitpod-io/workspace-images/issues/763), we'll be able to test in the workspace more quickly.

kylos101:
I am going to close this for now, as it is no longer needed.

PR #750, opened 2022-03-02
Get rid of PR assignment to Project

princerachit:
## Description

To have a single source of truth, all PRs should be continued to be
listed using the team label instead of adding it to team project board.

## Related Issue(s)

NA

## How to test

This can be only tested post merge.
The issue triage should continue to work and PRs should not get auto assigned to project after this is merged.

## Release Notes

```release-note
None
```

### DIFF
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -1,7 +1,7 @@
 name: Issue and PR Triage
 on:
   pull_request:
-    types: [opened, ready_for_review]
+    types: [opened]
   issues:
     types: [opened]
 jobs:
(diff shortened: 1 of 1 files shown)


princerachit:
>For pull request triage,in hindsight, if we auto assign the team: workspace label, that should fill the gap nicely. For example, that will cause related PRs to become [visible to the team when we query like so](https://github.com/pulls?q=is:open+is:pr+archived:false+label:%22team:+workspace%22+org:gitpod-io+sort:created-asc+draft:false).

👍🏾 

> The benefit of this approach will be, that we can remove the dependency on the graph QL query, I think.

Is it really important to get rid of GraphQL dependency? That is a recommended method of interacting with Projects (Beta).
We still want to continue adding the issues to our project to triage it further, right? So we still need GH Graphql APIs.

princerachit:
Please merge only after @kylos101 's review

kylos101 (approved):
LGTM

PR #754, opened 2022-03-02
Ensure docker compose subcommand is available

### LINKED ISSUE #577: Update Docker Compose to Version 2
See https://docs.docker.com/compose/cli-command/

Furisto:
## Description
The docker compose command (_not_ docker-compose) is not available. We already download the binary but it is not in the location that docker expects. See [installation instructions.](https://docs.docker.com/compose/cli-command/#install-on-linux).

## Related Issue(s)
Fixes #577

## How to test
- Run workspace image with docker tools installed
- Call docker compose -> The command should be available

## Release Notes
```release-note
Make docker compose subcommand available
```


sagor999 (approved):

kylos101 (approved):
:+1:

Issue #763, opened 2022-01-24
Allow building only desired chunks


kylos101:
## Is your feature request related to a problem? Please describe

Inspecting how a new test is running requires building all chunks and running all tests, which makes for a long dev loop.

## Describe the behaviour you'd like

Provide a means to build and run a subset of chunks (which could also be a single chunk)

## Describe alternatives you've considered

Create a new branch to do testing from main, just with the chunks I want to build and test.
Once the tests are working, copy them into yet another new branch created from main.

## Additional context

This should help folks more easily test, which will hopefully encourage more authoring of tests.

kylos101:
@princerachit , if you could take on this issue, that would help us get very close to completing this related epic. Let us know? 

@Furisto I'm not sure if you have any script prepared for this already? If you do, can you share?

PR #764, opened 2022-03-08
Allow building specific chunks locally

princerachit:
## Description

Update `dazzle.yaml` to take inputs from user to build chunks and create a combination.

## Related Issue(s)

Fixes https://github.com/gitpod-io/workspace-images/issues/763

## How to test

Run `.dazzle-up.sh`, you should see that all the chunks are built and all the combinations as specified in `dazzle.yaml` are built.

To build specific chunks and combine those chunks use something like this:
```console
./dazzle-up.sh  -c chunk1 -c chunk2 -n combo
``` 

a working example:

```console
./dazzle-up.sh  -c lang-c -c dep-cacert-update -n mychangecombo
```

## Release Notes

```release-note
None
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 /dazzle
 /legacy
+/.dazzle.yaml.orig
+/.dazzle.yaml.temp
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -6,4 +6,4 @@ ENV BUILDKIT_VERSION=0.9.3
 ENV BUILDKIT_FILENAME=buildkit-v${BUILDKIT_VERSION}.linux-amd64.tar.gz
 
 # Install custom tools, runtime, etc.
-RUN sudo su -c "cd /usr; curl -L https://github.com/moby/buildkit/releases/download/v${BUILDKIT_VERSION}/${BUILDKIT_FILENAME} | tar xvz"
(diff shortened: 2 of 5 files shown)


Furisto:
My [approach](https://github.com/gitpod-io/workspace-images/commit/454d2dd46246d15babe4931e7a048516d7bd5d0e) has been that you can specify a specific combination (e.g. python) and it will find all necessary chunks, build them and then do the combine. :wink:

Furisto:
@axonasif Would this still be an issue with #761? The combinations specify version specific chunks so this should not lead to building multiple variants. Can you give an example where this is still the case?

axonasif:
> @axonasif Would this still be an issue after #761?

It should partially solve the issue as I see that the extra go variants were removed.

> The combinations specify version specific chunks so this should not lead to building multiple variants. Can you give an example where this is still the case?

Yes but that's for `dazzle combine` command I assume, but when you do `dazzle build` before running the combine command it will build all the variants of a chunk and you can only `dazzle project ignore lang-go` but not `dazzle project ignore lang-go:version`. That's what I observed.

axonasif:
> This has now been added to dazzle.

Awesome, thank you! Will keep an eye for when a binary release is published with this patch.

Furisto:
@princerachit I think both of our approaches are valuable. I propose that I add my own implementation to the repository as well and we separate it into different scripts i.e. build-all, build-chunks and build-combination.

Furisto (approved):

PR #783, opened 2022-03-23
fix: /usr/games missing PATH entry

axonasif:
## Description
Packages like [`cowsay`](https://packages.ubuntu.com/bionic/all/cowsay/filelist) get installed at `/usr/games` but it's not on PATH. Although I'm not sure if what I did on this PR is the recommended approach, please suggest.

## Related Issue(s)

Fixes https://app.frontapp.com/open/cnv_ix5jz62

## Release Notes

```release-note
/usr/games is now added on PATH for packages like cowsay
```

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -58,3 +58,6 @@ RUN sudo echo "Running 'sudo' for Gitpod: success" && \
 
 # configure git-lfs
 RUN sudo git lfs install --system
+
+# Custom PATH additions
+ENV PATH=/usr/games:$PATH


utam0k:
@axonasif Can I ask you to write the `Release Notes` section?

utam0k:
I agree to add this path because the use of `/usr/games` is occasional. I would like to hear what others have to say.

princerachit:
This affects the base image. I am curious where this use case is coming from. How much beneficial is this for users?
I am inclined to add such changes in a single PR so that the base image is changed only once.

Furisto:
If we add /usr/games, should we also add /usr/local/games? I would argue no, because this location is typically managed by the user himself and not used by package managers. Therefore it should be the responsibility of the user to add it to the PATH when he adds a game. Just putting this to discussion so that as @princerachit mentioned we do not have to change the base image later again.

kylos101 (approved):
+1 to /usr/games (but not /usr/local/games).

LGTM

PR #784, opened 2022-03-30
An optional preload flag to control the YugabyteDB initialization

srinivasa-vasu:
Signed-off-by: Srinivasa Vasu <srinivasan.surprise@gmail.com>

## Description
An optional `.nopreload` file as part of the repo restricts the workspace from installation & config of the DB. By default, when the workspace launches, DB will be initialized and running. This optional file gives the flexibility to control the initialization behavior. With the `.nopreload` file, users can control the DB via the .gitpod.yml file.

### DIFF
--- a/chunks/tool-yugabytedb/Dockerfile
+++ b/chunks/tool-yugabytedb/Dockerfile
@@ -38,7 +38,7 @@ ENV YCQL_API_PORT=12000
 
 # re-initialization is automatically handled
 RUN echo "\n# yugabytedb start command" >> /home/gitpod/.bashrc.d/100-yugabyedb-launch
-RUN echo "yugabyted start --base_dir=$STORE --listen=$HOST" >> /home/gitpod/.bashrc.d/100-yugabyedb-launch
+RUN echo "[[ -f \${GITPOD_REPO_ROOT}/.nopreload ]] || yugabyted start --base_dir=$STORE --listen=$HOST > /dev/null" >> /home/gitpod/.bashrc.d/100-yugabyedb-launch
 
 RUN chmod +x /home/gitpod/.bashrc.d/100-yugabyedb-launch
 


utam0k:
@gitpod-io/engineering-workspace I'm not sure if we can use `.nopreload` in the image. At least I couldn't find it used in other cases. Can someone please review this?

srinivasa-vasu:
@utam0k - This is to give flexibility to developers. They can keep the .nopreload file as part of the source repo to control the bash script from initializing. It would allow them to define their custom configuration flexibly. Pls, refer to this => https://github.com/srinivasa-vasu/yb-git-pod/blob/main/.gitpod-cluster-config.yml. More of a flag to control the default convention. Overriding this will allow them to define their custom configuration.

chunks/tool-yugabytedb/Dockerfile
```diff
@@ -38,7 +38,7 @@ ENV YCQL_API_PORT=12000
 
 # re-initialization is automatically handled
 RUN echo "\n# yugabytedb start command" >> /home/gitpod/.bashrc.d/100-yugabyedb-launch
-RUN echo "yugabyted start --base_dir=$STORE --listen=$HOST" >> /home/gitpod/.bashrc.d/100-yugabyedb-launch
+RUN echo "[[ -f \${GITPOD_REPO_ROOT}/.nopreload ]] || yugabyted start --base_dir=$STORE --listen=$HOST > /dev/null" >> /home/gitpod/.bashrc.d/100-yugabyedb-launch
```
princerachit, line 41:
```suggestion
RUN echo "[[ ${DISABLE_AUTO_YB_START} == "true" ]] || yugabyted start --base_dir=$STORE --listen=$HOST > /dev/null" >> /home/gitpod/.bashrc.d/100-yugabyedb-launch
```
IMO using an environment variable for the check makes more sense compared to a file as it is a usual practice in docker world to enable or disable a service using env variables.

srinivasa-vasu: My understanding is env variables are terminal-specific in Gitpod unless you create a project (which may not be the case for everyone). When you launch a new terminal inside VSCode, it's going to get reset. To me .nopreload file in the source repo is similar to .gitignore. Including that will nullify the default convention.

utam0k: How many cases would you like to use this? Could you consider supporting only those with high usage and using a custom image for the others?

princerachit (approved):

PR #787, opened 2022-03-31
Bump Go to 1.18

csweichel:
## Description
Bumps Go to 1.18

### DIFF
--- a/chunks/lang-go/chunk.yaml
+++ b/chunks/lang-go/chunk.yaml
@@ -1,4 +1,4 @@
 variants:
-  - name: "1.17.5"
+  - name: "1.18.0"
     args:
-      GO_VERSION: 1.17.5
+      GO_VERSION: 1.18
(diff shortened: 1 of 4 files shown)


chunks/lang-go/chunk.yaml
```diff
@@ -2,3 +2,6 @@ variants:
   - name: "1.17.5"
```
Furisto, line 2: 1.17.5 is not referenced anywhere, so we should remove it.

csweichel: not sure - if we want to offer a "features" like feature, this would still be handy. We could also remove it now and add it back in later.

Furisto: Removing it now and adding it back later is the best way in my opinion. Leaving it in "just in case" will increase build times without providing value in return and we might not actually need it in the future.

csweichel: done

Furisto (approved):

PR #789, opened 2022-04-01
Add new images for different python versions

### LINKED ISSUE #600: Install newer versions of python in gitpod/workspace-full image
## Describe the behaviour you'd like

I'd like to be able to switch to a newer python version using pyenv without installing it myself.

## Describe alternatives you've considered

Custom Dockerfile or installing in the init step, but both of these solutions 1) require more work from your users, 2) make the prebuild take longer. In my experience, the init solution adds 1-2 minutes to the prebuild and the Dockerfile solution adds 5-15 minutes.

## Additional context

This should be as easy as adding

```
RUN pyenv install 3.9.7 && pyenv install <some other newer version>
```

to the gitpod/workspace-full Dockerfile.

To get even more up-to-date versions, I'd recommend installing pyenv using [Github checkout](https://github.com/pyenv/pyenv#basic-github-checkout) instead of the pyenv installer. So, replace

```
RUN curl -fsSL https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer
```

with

```
RUN git clone https://github.com/pyenv/pyenv.git ~/.pyenv
```

## How to implement

Add dazzle variants for Python 3.6, 3.7, 3.8, 3.9, and 3.10, and upload distinct workspace-images for each variant. This would be similar to what we do now with Java 11/17, and Node/Node LTS, except we'd have 5 Python variants.

Furisto:
## Description
Adds the currently supported versions of python plus the most recent not support one (3.6 - 3.10)

## Related Issue(s)
Fixes #600

## How to test
n.a.

## Notes for reviewers
I kept the current python image (without a version number) and referenced the python-3.8 chunk (same version as before). We should eventually phase out this image and only use versioned python images. Opinions?

## Release Notes
```release-note
Add images for supported python versions and most recent not supported one
```


sagor999 (approved):
>We should eventually phase out this image and only use versioned python images.

:+1:

PR #790, opened 2022-04-01
Autofix Python variants

jankeromnes:
## Description

Implements a local Autofix fixer for Python variants, based on https://github.com/autofix-dev/autofix/blob/master/fixers/upgrade-pyenv-tools.js

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


Furisto:
Superseded by #791 because Github does not get that it needs to merge into main now.

Issue #792, opened 2022-04-03
Make available base + minimum set of tools (no lang) image 


mruzekw:
I'm just playing around with custom docker images on gitpod and noticed that only the `gitpod/workspace-full` image provides tools like homebrew and nix.

The image is HUGE and takes a while to build.

It would be great to have an image without any lang runtimes so that we can just install what we'd like with homebrew or nix and have a lean(er) image to start with.

I'm just thinking something like base (dep-cacert-update, tool-docker, tool-tailscale) +

tool-brew
tool-chrome
tool-nix
tool-vnc

kylos101:
Hey @mruzekw , have you tried `workspace-base` as a base? I assume yes, because you references [its few chunks](https://github.com/gitpod-io/workspace-images/blob/8a4fcd02e7c09389af5e5e3e7860e3d69d82474a/dazzle.yaml#L3-L7). To recap, you can then [layer on your own changes in a dockerfile that you manage](https://www.gitpod.io/docs/config-docker#configure-a-custom-dockerfile:~:text=If%20you%20want%20a%20base%20image%20without%20the%20default%20tooling%20installed%20then%20use%20the%20gitpod/workspace%2Dbase%20image.).

Different people have different definitions of what it means to have a base image. Some might want Firefox, instead of Chrome, and not want Nix or VNC.

For the meantime, `workspace-base` is what you're looking for. For the long term, we'd look to rewrite our image builder, so that users can dynamically specify the chunks that they desire (and they would be lazy loaded on workspace start).

kylos101:
👋 @mruzekw I am going to close this issue for now, please let me know if you'd like me to reopen, and if so, how else we might be able to help you? 🙏

PR #795, opened 2022-04-04
Add nix image

emilytrau:
## Description

Adds a minimal image variant with only the nix tool. This is useful for projects that leverage nix entirely for dependency management.

## How to test

n/a

## Release Notes

```release-note
Add new nix image
```

## Documentation

n/a

### DIFF
--- a/.github/promote-images.yml
+++ b/.github/promote-images.yml
@@ -11,6 +11,7 @@
     workspace-full: "20.*"
     workspace-full-vnc: "20.*"
     workspace-go: "20.*"
+    workspace-nix: "20.*"
     workspace-node: "20.*"
     workspace-node-lts: "20.*"
     workspace-python: "20.*"
--- a/.github/sync-containers.yml
+++ b/.github/sync-containers.yml
@@ -8,6 +8,7 @@ sync:
     - full
     - full-vnc
     - go
+    - nix
     - node
     - node-lts
     - python
--- a/dazzle.yaml
+++ b/dazzle.yaml
@@ -45,6 +45,11 @@ combiner:
       - base
       chunks:
         - lang-go:1.18.0
+    - name: nix
+      ref:
+      - base
+      chunks:
+        - tool-nix:2.6.0
     - name: node
       ref:
       - base


utam0k (approved):
Thanks for your contribution!

utam0k:
@princerachit I wonder why it doesn't include yet. Do you know some reasons?

princerachit:
hey @emilytrau I am not aware how widely nix is used by our users. Do we have this as a request from user or just something you think is useful.

I am asking these questions because nix is already shipped with `workspace-full` which is cached on our nodes and starts pretty fast.


@utam0k we did not have this because we initially targeted on having existing images migrated and then add new images based on user request

emilytrau:
@princerachit In projects where nix is the only base dependency it doesn't make sense to bring the rest of the language runtimes as well. Even just speaking for myself I would have a lot of use for this to reduce workspace startup time.

Furisto:
#792 is a similar request. We could add an image which only contains the base image and package managers.

Furisto:
Hey @emilytrau, I updated this PR to also include the brew chunk. That way we have an image which users can use as a base with their preferred package manager.

kylos101 (changes requested):
Hi @emilytrau :wave: , 

Could you change the [combination name](https://github.com/gitpod-io/workspace-images/pull/795/files#diff-b8b5d73ca5ee13982e492c43c0198eda34749eab5c1f8b3621f6512ec5af6f8bR8) from `pkg` to `nix`?

Also, may we ask for your help with a couple things, so that this can be included with our CI/CD?
   * There are a couple spots in our CI that need to have this added added [[1](https://github.com/gitpod-io/workspace-images/blob/fb61836cd6673a54861de621d41154a781b530c6/.github/promote-images.yml#L21)][[2](https://github.com/gitpod-io/workspace-images/blob/fb61836cd6673a54861de621d41154a781b530c6/.github/sync-containers.yml#L18)]

Thank you so much! :pray:

emilytrau:
> Hey @emilytrau, I updated this PR to also include the brew chunk. That way we have an image which users can use as a base with their preferred package manager.

Hey @Furisto. In this case I see this image targetting use in nix-language projects, rather than as a toolkit of the package managers. This is similar to the other language-specific images. FYI Brew adds 700MB/30% to the total size. 

IMG_0033

kylos101 (approved):
Thanks for the contribution @emilytrau ! I agree, keeping `nix` by itself is the right call, especially for folks like a language.

PR #797, opened 2022-04-06
Add neovim to base image

andrew-farries:
## Description

Add [Neovim](https://neovim.io/) to the base image, along with [tmux](https://github.com/tmux/tmux/wiki) and [fzf](https://github.com/junegunn/fzf).

Gitpod currently provides a basic user experience for users who work by SSHing into a workspace. For users who want a better experience, the only option currently is to install extra tools via their `dotfiles` repo. However this makes workspace startup times unacceptably slow. By adding more tools commonly used by CLI-centric dev workflows to the base image we enable these users to get more out of Gitpod.

## How to test

```
./build-combo.sh full
docker pull localhost:5000/dazzle:full
docker run --rm -it localhost:5000/dazzle:full /bin/bash
$ nvim --version
$ tmux -V
$ fzf --version
```

## Release Notes

```release-note
Add `nvim`, `tmux` and `fzf` to the base image for all workspaces
```

## Documentation

No changes required.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ You need to have a gitpod account in order to seamlessly contribute to this repo
 
 ## Tools
 
-This repo is already configured with a [.gitpod.yml](.gitpod.yml) which installs and sets up all the required dependency.
+This repo is already configured with a [.gitpod.yml](.gitpod.yml) which installs and sets up all the required dependencies.
 It also starts the services that are required to build this repo in dedicated bash shells.
 
 Here is a list of dependencies and tools:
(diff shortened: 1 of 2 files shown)


csweichel:
Base image changes are very expensive. Producing good ssh experience is a rather specialised concern which we could either cover as a chunk, or even as an ide image. I don't think we should add it to the base though.

andrew-farries:
How about adding a `tool-neovim` chunk and adding it to the `full` combination? If so, I can close this and make a PR for that instead.

csweichel:
> How about adding a `tool-neovim` chunk and adding it to the `full` combination? If so, I can close this and make a PR for that instead.

If we wanted to include neovim adding a `tool-neovim` chunk for it would certainly be the way to go.

andrew-farries:
Closing in favour of adding a `tool-neovim` chunk as discussed.

PR #803, opened 2022-04-11
Disable sdkman self update feature using new flag

princerachit:
## Description

sdkman uses a new flag to control self update. This PR updates that flag to the sdkman config in order to avoid asking user if they want to update the sdkman version. This as a result also fixes failing main branch builds.

## Related Issue(s)

NA

## How to test

NA

## Release Notes

```release-note
None
```


Furisto:
@princerachit Why wasn't this a problem in the PR builds?

princerachit:
> @princerachit Why wasn't this a problem in the PR builds?

During PR build every chunk is built. As a result the sdkman version installed in the chunk is latest.

When combine is performed on merges to main branch, existing java chunk is pulled from GAR. This is because there is no change in the java dockerfile and sdkman is not pinned to a version.

sagor999 (approved):

PR #806, opened 2022-04-21
Update go to v1.18.1

aledbf:
## Release Notes
```release-note
Update go to v1.18.1
```

### DIFF
--- a/chunks/lang-go/chunk.yaml
+++ b/chunks/lang-go/chunk.yaml
@@ -2,3 +2,6 @@ variants:
   - name: "1.18.0"
     args:
       GO_VERSION: 1.18
+  - name: "1.18.1"
+    args:
+      GO_VERSION: 1.18.1
--- a/dazzle.yaml
+++ b/dazzle.yaml
@@ -26,7 +26,7 @@ combiner:
       chunks:
         - lang-c
         - lang-clojure
-        - lang-go:1.18.0
+        - lang-go:1.18.1
         - lang-java:11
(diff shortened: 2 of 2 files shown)


chunks/lang-go/chunk.yaml
```diff
@@ -2,3 +2,6 @@ variants:
   - name: "1.18.0"
```
princerachit, line 2: Let us get rid of 1.18.0 as this won't be published with the change any more

aledbf: I disagree. We should be able to select versions. Getting rid of a chunk removes that option.

princerachit: Maybe I am misunderstanding this, but we are not publishing this variant anywhere after this change. In past the variants piled up quite a bit causing frequent timeouts during build which we cleaned up https://github.com/gitpod-io/workspace-images/pull/761/files

kylos101: @princerachit the new chunk will be done in `dazzle build`, which is generally quick.

We'll only use one of the two `go` chunk, 1.18.1, when we `dazzle combine`. Since the # of combinations isn't increasing, we will be less likely to timeout. If this were a new combination - I'd be worried - but it isn't, so I think we're okay.

Even if we do have a timeout, we can just run the build again (short term hack), and it'll more likely work.

kylos101 (approved):
:+1:

PR #808, opened 2022-04-21
refactor(lang-rust): Optimize the image and fix `rustup toolchain install` issue

axonasif:
## Description

`rustup toolchain install <toolchain>` command was conflicting with the wrapper script of it.
Also, some tiny optimizations was done. Now this image consumes `214MB` less space as we get rid of the unnecessary `$HOME/.cargo/registry` 🥳

## Related Issue(s)

Fixes #[this-on-discord](https://discord.com/channels/816244985187008514/966633482619084800/966633485156626442), reported by @jcornaz

## How to test

- Run a docker container of the `lang-rust`(workspace-rust) chunk.
- Run `rustup toolchain install stable nightly`
- Run `cargo new /tmp/foo` to check cargo is properly functioning.

## Release Notes

```release-note
```

### DIFF
--- a/tests/lang-rust.yaml
+++ b/tests/lang-rust.yaml
@@ -4,8 +4,10 @@
   assert:
   - status == 0
   - stdout.indexOf("rustc") != -1
-- desc: it should have properly configured rustup
-  command: [rustup show]
+- desc: it should have properly functioning rustup
+  command: [
+      rustup show && rustup update && rustup toolchain install nightly && cargo new /tmp/foo,
+    ] # Testing out `cargo` at the end is necessary to assert that rustup isn't messing things up
   entrypoint: [bash, -i, -c]
   assert:
   - status == 0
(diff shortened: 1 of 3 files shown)


tests/lang-rust.yaml
```diff
@@ -4,8 +4,8 @@
   assert:
   - status == 0
   - stdout.indexOf("rustc") != -1
-- desc: it should have properly configured rustup
-  command: [rustup show]
+- desc: it should have properly functioning rustup
+  command: [rustup show && rustup update && rustup toolchain install nightly && cargo new /tmp/foo] # Testing out `cargo` at the end is necessary to know that rustup didn't mess things up
```
sagor999, line 8: `rustup toolchain install nightly`
just double checking, do we indeed want to install nightly?

jcornaz: The bug was observable only after installing another toolchain. So, yes, to have a test that reproduces the problem it is necessary to install an additional toolchain such as `beta` or `nightly`.  `nightly` is relevant I think because it is the most used after `stable`.

sagor999 (approved):

PR #809, opened 2022-04-22
Trigger tailscale chunk rebuild

sagor999:
## Description

Trigger rebuild so that it can grab new pub key for that repo. 

## Related Issue(s)

Fixes #

## Release Notes

```release-note
none
```

PR #812, opened 2022-04-22
[tailscale] Fix tailscale chunk

csweichel:
## Description
Adding PGP keys to apt does not work with dazzle.
This change removes the the source list to avoid this problem in subsequent `apt-get` operations.

### DIFF
--- a/chunks/tool-tailscale/Dockerfile
+++ b/chunks/tool-tailscale/Dockerfile
@@ -9,6 +9,10 @@ ENV TRIGGER_REBUILD=1
 RUN curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/focal.gpg | sudo apt-key add - \
     && curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/focal.list | sudo tee /etc/apt/sources.list.d/tailscale.list \
     && apt-get update \
-    && apt-get install -y tailscale
+    && apt-get install -y tailscale \
+    && rm /etc/apt/sources.list.d/tailscale.list
+
+## "apt-key add" adds a PGP to a joint file which does not make it through dazzle.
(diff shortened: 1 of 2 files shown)


utam0k (approved):

chunks/tool-tailscale/Dockerfile
```diff
@@ -9,6 +9,10 @@ ENV TRIGGER_REBUILD=1
 RUN curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/focal.gpg | sudo apt-key add - \
```
princerachit, line 9: Should we change the way we store apt keys in this PR itself?

https://github.com/gitpod-io/workspace-images/blob/main/CONTRIBUTING.md#adding-a-chunk

```
# fetch keyring over https connection and unpack it using gpg's --dearmor option
curl -fsSL https://apt.my-secure.org/my-unofficial-repo.gpg.key | sudo gpg --dearmor -o /usr/share/keyrings/my-unofficial-repo.gpg.key
# and then add them to a apt key sources list.
echo "deb [signed-by=/usr/share/keyrings/my-unofficial-repo.gpg.key] http://apt.my-secure.org/focal/ \
my-unofficial-repo-toolchain main" | sudo tee /etc/apt/sources.list.d/my-unofficial-repo.list > /dev/null
```

csweichel: That makes sense - @princerachit please feel free to push a commit or open a new PR

princerachit:
This PRs build will likely fail, the fix is part of another PR, see [commit](https://github.com/gitpod-io/workspace-images/pull/809/commits/c5b2949964139c8bf9333906a95d647443158f7d)

kylos101 (approved):
:+1:

PR #765, opened 2022-03-08
fix(lang-rust): Custom dockerfile issue for cargo-install to non-existent /workspace dir

axonasif:
## Description

CARGO specific env var modifications are now done inside workspace-session but not for docker env.

## Related Issue(s)

Fixes [this on discord](https://canary.discord.com/channels/816244985187008514/946731644465852466/946805048116121640)

## Release Notes

```release-note
```

PR #818, opened 2022-04-26
fix(chunks/lang-go): Be context aware of /workspace

axonasif:
## Description

Please check #765 for the reason of this change 🙏 

## Release Notes

```release-note
```

## Documentation

Related: https://github.com/gitpod-io/website/pull/1947


utam0k:
Is there any way to test to make sure?

jenting:
The issue you linked is related to Rust however this PR address to Go, is this change across all languages?

axonasif:
> The issue you linked is related to Rust however this PR address to Go, is this change across all languages?

If you install something with `go` cli from your custom dockerfile, it will try to install them under `/workspace/go` but that dir does not exist in the imagebuild environment and even if you create a `/workspace` dir, it will get overlaid due to `mount` on workspace start.

The linked issue was resolving the same kind of problem but for rust, this is for go.

axonasif:
> Is there any way to test to make sure?

Yes but I'm not sure if that is necessary since we are already using the `go` CLI inside this chunk's dockerfile, we're not affecting the docker environment but the workspace-session when a user starts a workspace.

princerachit (approved):

PR #825, opened 2022-05-01
Add `tmux` and `neovim` chunks and use them in the `full` image

andrew-farries:
## Description

Add [neovim](https://neovim.io/) and [tmux](https://github.com/tmux/tmux/wiki) chunks and include them in the `full` workspace image.

We've recently done some great work to make Gitpod easier to use for users with CLI based workflows by making workspaces simpler to access via SSH and supporting `dotfiles` repos to allow customisation of the CLI environment. 
As this is a use-case we want to support, we should also have tools available in the default workspace image that support such a workflow. 

This PR adds `tmux` and `neovim` to the `full` image. `tmux` is the terminal multiplexer, and `neovim` is a modern fork of `vim` with full LSP support that serves as an alternative to full-blown IDEs for users who prefer the CLI.

## How to test

Dazzle tests added for the two new tools.

## Release Notes

```release-note
Add `tmux` and `neovim` tools to the `full` workspace image
```

## Documentation

None required AFAIK.

### DIFF
--- a/chunks/tool-neovim/Dockerfile
+++ b/chunks/tool-neovim/Dockerfile
@@ -1,0 +1,11 @@
+ARG base
+FROM ${base}
+
+USER gitpod
+
+# Dazzle does not rebuild a layer until one of its lines are changed. Increase this counter to rebuild this layer.
+ENV TRIGGER_REBUILD=1
+
+RUN curl -o /tmp/nvim.deb -fsSL https://github.com/neovim/neovim/releases/download/v0.7.0/nvim-linux64.deb \
+    && sudo apt install /tmp/nvim.deb \
+    && rm -f /tmp/nvim.deb
--- a/chunks/tool-tmux/Dockerfile
+++ b/chunks/tool-tmux/Dockerfile
@@ -1,0 +1,9 @@
+ARG base
+FROM ${base}
+
+USER gitpod
+
+# Dazzle does not rebuild a layer until one of its lines are changed. Increase this counter to rebuild this layer.
+ENV TRIGGER_REBUILD=1
+
+RUN sudo install-packages tmux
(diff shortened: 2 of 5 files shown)


utam0k:
Hi, @andrew-farries. Thanks for your PR. What about preparing an image for cli instead of installing it in full image? I think most users use vscode. In that case, I feel that these two alternative functions are provided in vscode.

utam0k:
@gitpod-io/engineering-workspace I'd like to see what others have to say?

csweichel:
> Hi, @andrew-farries. Thanks for your PR. What about preparing an image for cli instead of installing it in full image? I think most users use vscode. In that case, I feel that these two alternative functions are provided in vscode.

Totally follow that sentiment.
From a user-experience perspective that would not be great though. The workspace image is set on a project level, using SSH is a personal preference.

The right way to go would be to introduce an "SSH desktop IDE image" (cc @loujaybee @akosyakov). For now though I reckon we could merge this change.

akosyakov:
> The right way to go would be to introduce an "SSH desktop IDE image" (cc @loujaybee @akosyakov). For now though I reckon we could merge this change.

👍 

@loujaybee We track it right now in https://github.com/gitpod-io/gitpod/issues/9323 but for VIM, so maybe we rather turn it into more generic. I also it can really quick win, having something like vim, tmux, neovim in it will be enough.

akosyakov:
Actually @andrew-farries you can ahead and give it a try by opening a PR:
- which contributes CLI image under https://github.com/gitpod-io/gitpod/tree/main/components/ide, i.e. `ide/cli`. You can have a look at JetBrains and VS Code images next to it to learn how.
- add a new configuration in installer for CLI users https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/components/server/ide/configmap.go

After that you can ping us to update the start screen a bit to provide you with copy/paste CLI command to start.

andrew-farries:
@akosyakov Thanks;  I'll attempt a PR to add a new IDE image and close this.

PR #844, opened 2022-05-25
Fix skopeo login command

princerachit:
## Description

Fix skopeo login command

## Related Issue(s)

Job fails https://github.com/gitpod-io/workspace-images/runs/6579934057?check_suite_focus=true

## How to test

NA

## Release Notes

```release-note
None
```


jenting:
The root cause probably is because we install different skopeo version:
- [push-main.yaml](https://github.com/gitpod-io/workspace-images/runs/6564313113?check_suite_focus=true) with version `100:1.3.0-1`.
- [dockerhub-release.yml](https://github.com/gitpod-io/workspace-images/runs/6579934057?check_suite_focus=true) with version `100:1.5.0-2`.

PR #868, opened 2022-06-20
fix(lang-python): pyenv and pip not working as expected between prebuilds and workspace reboots

axonasif:
## Description

Python wasn't properly functioning for prebuilds and between-reboots for Gitpod as many users expected and this has been around for some time.
And we also auto-generate a `.gitpod.yml` for python projects with the following content:
```yaml
tasks:
  - init: pip install -r requirements.txt
```
This causes a bad user experience and they have to figure out that they'll need to swap `init` with `before` or run the command again manually. But that's not what we intended it to be like.

## Related Issue(s)

Fixes:
- [x] `pip install x`/`pyenv <command>` has no effect when done from `init` prebuild
- [x] python packages and global `pyenv` changes does not persist between workspace reboots
- [ ] `pip install --user` has no effect when run from custom dockerfile

## How to test

- [ ] TODO
- [ ] Update the `test/lang-python.yaml`, add new tests.

## Release Notes

```release-note
pip and pyenv functions as expected on workspace-python
```


sagor999:
@axonasif #875 is merged. Can you rebase please?

jenting:
@axonasif could you rebase it, please? 🙏

sagor999:
Rebased it.

sagor999 (approved):

ccll:
Unfortunately `pyenv virtualenv ...` still does not work, may we also have this supported?

My project have multiple python microservices and each has it's own virtualenv, and I'm heavily using the `pyenv virtualenv ...` and `pyenv local ...` to automate the switch of all these virtualenvs.

axonasif:
Okay I did not cover that part as well, have to explore, thanks for reporting @ccll.

I gave it a quick try, could you please elaborate on what issues you're facing with it?
Any example repository where I can reproduce your issue would be super helpful.

PR #879, opened 2022-07-15
Install latest git version

aledbf:
Force build of base image to update git to latest version

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,8 +1,8 @@
 FROM gitpod/workspace-full
 
-ENV RETRIGGER=2
+ENV RETRIGGER=3
 
-ENV BUILDKIT_VERSION=0.10.2
+ENV BUILDKIT_VERSION=0.10.3
 ENV BUILDKIT_FILENAME=buildkit-v${BUILDKIT_VERSION}.linux-amd64.tar.gz
 
 USER root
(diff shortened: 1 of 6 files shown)


.gitpod.Dockerfile
```diff
@@ -1,16 +1,16 @@
 FROM gitpod/workspace-full
 
-ENV RETRIGGER=2
+ENV RETRIGGER=3
 
-ENV BUILDKIT_VERSION=0.10.2
+ENV BUILDKIT_VERSION=0.10.3
```
kylos101, line 5: Should we also update in dazzle and gitpod repos? We usually do those, and then update the dazzle and buildkit here in workspace-images. :thinking: 

I am not sure if the [performance regression](https://github.com/moby/buildkit/releases/tag/v0.10.3) that they mention will help improve dazzle build/combine times, though.

aledbf: We've been using 0.10.3 since May https://github.com/gitpod-io/gitpod/pull/9881

kylos101: I gotcha. I was thinking dazzle [[1](https://github.com/gitpod-io/dazzle/blob/main/go.mod#L20)], and assumed (doh!) that Gitpod was not done if dazzle was not done.

sagor999 (approved):

PR #887, opened 2022-07-20
feat: enable prebuilds via persisting bazel/blaze cache

ghuntley:
## Description

Bazel/Blaze by default stores its build cache in a place where Gitpod does not persist thus meaning Blaze/Bazel does not work with prebuilds.

## Related Issue(s)

L10 of https://github.com/carbon-language/carbon-lang/compare/trunk...gitpod-forks:carbon-lang:devenv/gitpod?expand=1#diff-454eff0561fdbb1cdf08bef7a76ddf0f82c65fca08a150c43650f5098ecf7118R10 should not be required.

## How to test

Tests are included but I have not tested running a workspace and am raising pull-request to trigger a CI job.

## Release Notes

```release-note
feat: enable prebuilds via persisting bazel/blaze cache
```

### DIFF
--- a/chunks/tool-bazel/Dockerfile
+++ b/chunks/tool-bazel/Dockerfile
@@ -1,0 +1,11 @@
+ARG base
+FROM ${base}
+
+USER gitpod
+
+# Dazzle does not rebuild a layer until one of its lines are changed. Increase this counter to rebuild this layer.
+ENV TRIGGER_REBUILD=1
+
+# store the blaze/bazel cache in a place where it will be persisted
+# See also https://docs.bazel.build/versions/main/output_directories.html
+RUN mkdir -p /home/gitpod/.cache/bazel && ln -s /home/gitpod/.cache/bazel /workspace/.bazel
--- a/tests/tool-bazel.yaml
+++ b/tests/tool-bazel.yaml
@@ -1,0 +1,5 @@
+- desc: bazel cache should be persisted to /workspaces
+  command: [file, /workspace/.bazel]
+  assert:
+  - status == 0
+  - stdout.indexOf("/workspace/.bazel: symbolic link to /home/gitpod/.cache/bazel") != -1
--- a/dazzle.yaml
+++ b/dazzle.yaml
@@ -32,6 +32,7 @@ combiner:
         - lang-python:3.8
         - lang-ruby:3.1
         - lang-rust
+        - tool-bazel
         - tool-brew
         - tool-nginx
         - tool-nix:2.6.0


tests/tool-bazel.yaml
```diff
@@ -0,0 +1,5 @@
+- desc: bazel cache should be persisted to /workspaces
+  command: [file, /workspace/.bazel]
+  assert:
+  - status == 0
+  - stdout.indexOf("/workspace/.bazel: symbolic link to /home/gitpod/.cache/bazel") != -1
```
Siddhant-K-code, line 5: I think it should be something like this: [🔨 `Dazzle build`](https://github.com/gitpod-io/workspace-images/runs/7421114218?check_suite_focus=true#step:10:5) step is failing

```suggestion
  - stdout.indexOf("symbolic link to /home/gitpod/.cache/bazel") != -1
```

Siddhant-K-code: OR

```suggestion
  - stdout.indexOf("/workspace/.bazel") != -1
```

chunks/tool-bazel/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+FROM ${base}
+
+USER gitpod
+
+# Dazzle does not rebuild a layer until one of its lines are changed. Increase this counter to rebuild this layer.
+ENV TRIGGER_REBUILD=1
+
+# store the blaze/bazel cache in a place where it will be persisted
+# See also https://docs.bazel.build/versions/main/output_directories.html
+RUN mkdir -p /home/gitpod/.cache/bazel && ln -s /home/gitpod/.cache/bazel /workspace/.bazel
```
axonasif, line 11: This change will be not be effective as this `/workspace` dir will be overlaid via a mount once a workspace session will be started by Gitpod.

It needs to be done from something like `.bashrc` with a conditional check for `GITPOD_REPO_ROOT`'s existence.
```suggestion
RUN RUN ts="$HOME/.bashrc.d/700-bazel" && printf '%s\n' \
						't="$HOME/.cache/bazel"' \
						's="${t//$HOME/\/workspace}"' \
						'test ! -L "$t" && rm -rf "$t" && mkdir -p "$s" && ln -s "$s" "$t"' \
						"rm -f $ts" > "$ts"
```

While there is an env var which we could set in such way too via bash as it's stated here: https://bazel.build/docs/output_directories#layout

But then I came across [this](https://stackoverflow.com/questions/40775920/how-can-i-make-bazel-use-external-storage-when-building#comment98156737_40775921) comment which got me thinking, so I guess we should still continue with the symlink method unless that comment is proven otherwise (I'm not familiar with `bazel`, so no idea)

tests/tool-bazel.yaml
```diff
@@ -0,0 +1,5 @@
+- desc: bazel cache should be persisted to /workspaces
+  command: [file, /workspace/.bazel]
```
axonasif, line 2: In case my other `suggestion` is considered, this one below would then be necessary too for emulating a workspace-session behavior.
```suggestion
  entrypoint: [env, GITPOD_REPO_ROOT=/workspace, bash, -ci]
  command: [file, /workspace/.bazel]
```

jenting:
It needs rebase to main branch.

jenting:
I rebase it.

kylos101:
> Tests are included but I have not tested running a workspace and am raising pull-request to trigger a CI job.

Hi @ghuntley :wave: , 

To test, I recommend:
1. Opening a Gitpod from this PR
2. `./build-combo.sh full` <-- this will take a bit, and is caching in your workspace
3. push the image from your workspace's local registry to Docker Hub
4. set the image (now on docker hub) as a base image for your repo where you want to test prebuilds
5. test prebuilds

kylos101 (changes requested):
There are some thoughts, to summarize:
1. testing
2. see suggestions from @axonasif and @Siddhant-K-code

PR #888, opened 2022-07-21
Experiment with 2 Java versions in the full image

### LINKED ISSUE #874: Java 17 for workspace-full
## Is your feature request related to a problem? Please describe
redhat.java pushed an update requiring java 17 be set up in the homepath. The default jdk is 11, there for the main java extensions do not work unless java 17 is installed.

## Describe the behaviour you'd like
I would recommend moving workspace-full to default at java 17 due to the vscjava.vscode-java-pack provided by redhat vscjava

## Describe alternatives you've considered
Manually configuring the home path in each docker file with java 17 installed. Increases time for start up. 

## Additional context
Extension Notification: "Activating extension 'redhat.java' failed: Java 17 or more recent is required to run the Java extension. Please download and install a recent JDK. You can still compile your projects with older JDKs by configuring ['java.configuration.runtimes'](https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements#java.configuration.runtimes)."

filiptronicek:
## Description
Add Java 17 to the mix to fix the [Java extension](https://marketplace.visualstudio.com/items?itemName=redhat.java) in VS Code Browser.

## Related Issue(s)

Fixes #874 
Fixes https://github.com/gitpod-io/website/issues/2365

## How to test
I am not sure how to test this change effectively. @axonasif could you help here please? 🙏

## Release Notes

```release-note
NONE
```


akosyakov:
I'm not sure it is how it should be done. We need Java 11 as primary, but add Java 17 as secondary. @kylos101 Who can support with it? 🙏

akosyakov:
I think we should try to add another line here: https://github.com/gitpod-io/workspace-images/blob/9717910880670c8914ed7b1d62d25575fc30a737/chunks/lang-java/Dockerfile#L15 to install java 17 or maybe GA I'm not sure what java vscode extension is following now.

filiptronicek:
> I think we should try to add another line here: 

That would presumably work, but I am not sure about it. It would mean that when users specify they only want Java 11, Java 17 would be in their image as well. I am thinking maybe a third option (with the two current ones being for Java 11 and 17), which is something like `11+17`, which would download both versions and set 11 as default.

akosyakov:
> That would presumably work, but I am not sure about it. It would mean that when users specify they only want Java 11, Java 17 would be in their image as well. I am thinking maybe a third option (with the two current ones being for Java 11 and 17), which is something like 11+17, which would download both versions and set 11 as default.

It is temporary hack to fix Java extension. Even if a user specifies Java 11, we have to put Java 17, otheriwse VS Code Java is not working right now.

filiptronicek:
Gave this another go and the only change I did is that I forced Java 17 to install in any Java image you choose as a user, but never make it the default. Sounds good, @akosyakov?

chunks/lang-java/Dockerfile
```diff
@@ -2,6 +2,7 @@ ARG base
 FROM ${base}
 
 ARG JAVA_VERSION
+ARG JAVA_VERSION_SECONDARY
```
akosyakov, line 5: Leftover?

filiptronicek: Oh yes, thanks for noting. I removed it as well, but forgot to push 😭

filiptronicek: https://github.com/gitpod-io/workspace-images/pull/888/commits/946522bde06dad4ffa7eed7ca01c4a2d99fbd458

andreafalzetti (approved):
Tested in https://github.com/gitpod-io/spring-petclinic.

I've published the image to a tmp dockerhub repo, and used that as workspace image for petclinic. The error notification is gone, but we get a warn to reload for some reason.

The webview showing the prompt to install java 17 is also gone, so it looks like it will work as we expect.

### Before
Screenshot 2022-07-26 at 14 52 54

###  After
Screenshot 2022-07-26 at 14 52 33

jenting:
I'm 🙆‍♂️ with this change, but I'd like @akosyakov to decide whether to merge it or not.

akosyakov:
I think it is good till we have proper fix in Open VSX

chunks/lang-java/Dockerfile
```diff
@@ -8,18 +8,18 @@ USER gitpod
...
-             && printf '<settings>\n  <localRepository>/workspace/m2-repository/</localRepository>\n</settings>\n' > /home/gitpod/.m2/settings.xml \
-             && echo 'export SDKMAN_DIR=\"/home/gitpod/.sdkman\"' >> /home/gitpod/.bashrc.d/99-java \
-             && echo '[[ -s \"/home/gitpod/.sdkman/bin/sdkman-init.sh\" ]] && source \"/home/gitpod/.sdkman/bin/sdkman-init.sh\"' >> /home/gitpod/.bashrc.d/99-java"
-# above, we are adding the sdkman init to .bashrc (executing sdkman-init.sh does that), because one is executed on interactive shells, the other for non-interactive shells (e.g. plugin-host)
-ENV GRADLE_USER_HOME=/workspace/.gradle/
+	&& source "$HOME/.sdkman/bin/sdkman-init.sh" \
+	&& sed -i -e 's/sdkman_selfupdate_enable=true/sdkman_selfupdate_enable=false/g' \
+	-e 's/sdkman_selfupdate_feature=true/sdkman_selfupdate_feature=false/g' "$HOME/.sdkman/etc/config" \
+	&& sdk install java ${JAVA_VERSION} \
+	&& sdk install java 17.0.3.fx-zulu \
```
kylos101, line 17: We support two Java chunk variants (11 and 17):
https://github.com/gitpod-io/workspace-images/blob/main/chunks/lang-java/chunk.yaml

We use the 11 chunk in workspace-full:
https://github.com/gitpod-io/workspace-images/blob/8b5d017c319bb8364d2ab327500b505d80686444/dazzle.yaml#L30, which lands here https://hub.docker.com/r/gitpod/workspace-full

We also use it in a dedicated image for 11:
https://github.com/gitpod-io/workspace-images/blob/8b5d017c319bb8364d2ab327500b505d80686444/dazzle.yaml#L139, which lands here https://hub.docker.com/r/gitpod/workspace-java-11

As you can see we use a separate 17 chunk to assemble the 17 combination:
https://github.com/gitpod-io/workspace-images/blob/8b5d017c319bb8364d2ab327500b505d80686444/dazzle.yaml#L144, which lands here https://hub.docker.com/r/gitpod/workspace-java-17

The change on line 17 is hard coding the Java variant, I get why (because you want both in the same image).

However, this will also add Java 17 to the Java 11 image: https://hub.docker.com/r/gitpod/workspace-java-11

In other words, this feels like a breaking change to me...I think we need more detailed testing, to ensure we're not breaking `workspace-full`, `workspace-java-11`, and `workspace-java-17`.

kylos101: In hindsight, it might make more sense to create a new, separate chunk, that we only use in workspace-full. We would call it "two chunk java", or something like that, and only use it in workspace-full.

kylos101 (changes requested):
Is Java 17 required, even if the user wants to use our exclusive Java 11 image, separate from workspace-full?

It would be great if more test results could be shared. For example, the one repo was tested with Java 1.8, but, it may be good to test other versions, too.

Also, there are some dazzle fundamentals being ignored (by including two types of Java in the same chunk), and then reusing it across different images, rather than use related variants.

akosyakov:
> This might be okay, but it warrants pausing to think about it. I need to look at this again with a fresh pair of eyes.

VS Code Java requires now Java 17. For one month users having bad experience requiring them to install it manually for each our image and configure. We don't want to change a java version used for user apps, i.e. 8 and 11. That's why we add Java 17 next to it not as default but only fix VS Code Java extension. It is also a temporary solution, till we can get OpenVSX deploy new features which supports bundling of JDK in VS Code extension then we can remove it.

I'm bit frustrated that already like 6-7 people look at it and it cannot be landed for a month and noone can land :(

chunks/lang-java/Dockerfile
akosyakov, line 17: Why this PR has so many changes I thought only one line is required: `sdk install java 17.0.3.fx-zulu`?

akosyakov: > However, this will also add Java 17 to the Java 11 image: https://hub.docker.com/r/gitpod/workspace-java-11

it is intended, without that Java in VS Code experience is bad now

axonasif: It was me, I had made some commits but it looks like that they were squashed together into the first commit and then force pushed 😅 :
Screenshot 2022-07-27 at 5 38 24 PM

Do you want me to push these for the sake of commit history?
Screenshot 2022-07-27 at 5 50 14 PM

andreafalzetti: > It was me, I had made some commits but it looks like that they were squashed together into the first commit and then force pushed 😅 : Screenshot 2022-07-27 at 5 38 24 PM
> 
> Do you want me to push these for the sake of commit history? Edit: I merged and pushed. Screenshot 2022-07-27 at 5 50 14 PM

@akosyakov and I agreed to rebase the PR (as we do for gitpod's main repo).

axonasif: > @akosyakov and I agreed to rebase the PR and squash the 16 commits to avoid noise (as we do for gitpod's main repo).

Fair enough 👍 
I think we can use the `Squash and merge` button from here on GitHub, that does the same thing a bit differently right? Since you know, force pushing can cause issues on local branches when multiple people are collaborating on the same PR 🙏 
Just my two cents.

akosyakov: Can we revert please all noise to the minimal one line change? It should have been very simple and quick PR to unblock our Java users.

axonasif: Sure 👍 
I reverted the extra changes. While adding the new tests I had faced an issue, so I thought to put the fix in here an an one-off thing 😅 
Will create a separate PR for that later 👍 
Sorry for the trouble 🙏

kylos101: Thanks @akosyakov , that helps me understand what we're trying to accomplish. Let me look at this PR again, to see how testing was done, to see if I have any other questions. I looked last night, and it sounded like only one image was tested.

kylos101: Thank you @akosyakov and @axonasif , this is helpful.

kylos101:
@andreafalzetti [which image did you test](https://github.com/gitpod-io/workspace-images/pull/888#pullrequestreview-1050963649)? I checked [here](https://github.com/gitpod-io/workspace-images/pull/888#pullrequestreview-1050963649), and it looks like you tested with one image, I assume `workspace-full`, but, I didn't see mentions of `workspace-java-11` or `workspace-java-17`.

kylos101:
@akosyakov once [the testing is validated](https://github.com/gitpod-io/workspace-images/pull/888#issuecomment-1197192703), I'm happy to continue. Thanks a lot for pushing to simplify the change.

kylos101:
Hey, I talked with @andreafalzetti , and he's going to rebuild using dazzle, the 11 and 17 images in a Gitpod, publish to Docker, and test with those. For background, he build using Docker, rather than use Dazzle. Now, [he's gonna use this](https://github.com/gitpod-io/workspace-images/blob/main/CONTRIBUTING.md#build-specific-combination).

andreafalzetti:
@kylos101 I did a second round of testing using the approach that we agreed upon, I've recorded a 4min loom to show what I've done: https://www.loom.com/share/f93ed5e8955c4ba0a1a7cdd3bdce237d 🙏 

cc @akosyakov

akosyakov:
@kylos101 Could you have a look please again? Can we land it?

kylos101 (approved):
Thanks for rebuilding with dazzle, @andreafalzetti , and sharing the test results!

kylos101:
@akosyakov this is merged, and building now. It'll be available as the `latest` tag next week on Tuesday as part of the typical deploy. If there are any users that need sooner, they can use the timestamp tag after the build finishes (which'll be ready in < 2 hours).

Issue #639, opened 2022-01-23
Add yarn global bin dir to PATH


thisisommore:
It will be nice if we have yarn global bin dir in the PATH
yarn is being installed here https://github.com/gitpod-io/workspace-images/blob/a8cf9bb750cd934f6001a1d90e945c4528fd4623/full/Dockerfile#L130
but while setting path only for node bin is set
https://github.com/gitpod-io/workspace-images/blob/a8cf9bb750cd934f6001a1d90e945c4528fd4623/full/Dockerfile#L134
![image](https://user-images.githubusercontent.com/51229945/150689715-dbb95e2b-2dcf-4d6a-bf39-dff4aaf6b5cd.png)

PR #907, opened 2022-08-22
fix: package manager bugs for Node.js

gengjiawen:
## Related Issue(s)

Fixes https://github.com/gitpod-io/workspace-images/issues/639
Fixes https://github.com/gitpod-io/workspace-images/issues/913

Also fix pnpm setup issue.

## How to test

build and test works.
```
docker build -t test --build-arg base=gitpod/workspace-base --build-arg NODE_VERSION=18.7.0 
```
I also add test cases in yml too.

## Release Notes

```release-note
```

### DIFF
--- a/chunks/lang-node/Dockerfile
+++ b/chunks/lang-node/Dockerfile
@@ -16,8 +16,7 @@ RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh |
     && bash -c ". .nvm/nvm.sh \
         && nvm install v${NODE_VERSION} \
         && nvm alias default v${NODE_VERSION} \
-        && npm install -g yarn \
-        && yarn global add typescript yarn pnpm node-gyp" \
+        && npm install -g typescript yarn pnpm node-gyp" \
     && echo ". ~/.nvm/nvm-lazy.sh"  >> /home/gitpod/.bashrc.d/50-node
 # above, we are adding the lazy nvm init to .bashrc, because one is executed on interactive shells, the other for non-interactive shells (e.g. plugin-host)
 COPY --chown=gitpod:gitpod nvm-lazy.sh /home/gitpod/.nvm/nvm-lazy.sh
--- a/chunks/lang-node/Dockerfile
+++ b/chunks/lang-node/Dockerfile
@@ -9,12 +9,15 @@ USER gitpod
 ENV TRIGGER_REBUILD=1
 ENV NODE_VERSION=${NODE_VERSION}
 
+ENV PNPM_HOME=/home/gitpod/.pnpm
+ENV PATH=/home/gitpod/.nvm/versions/node/v${NODE_VERSION}/bin:/home/gitpod/.yarn/bin:${PNPM_HOME}:$PATH
+
 RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | PROFILE=/dev/null bash \
     && bash -c ". .nvm/nvm.sh \
         && nvm install v${NODE_VERSION} \
         && nvm alias default v${NODE_VERSION} \
-        && npm install -g typescript yarn pnpm node-gyp" \
+        && npm install -g yarn \
+        && yarn global add typescript yarn pnpm node-gyp" \
     && echo ". ~/.nvm/nvm-lazy.sh"  >> /home/gitpod/.bashrc.d/50-node
 # above, we are adding the lazy nvm init to .bashrc, because one is executed on interactive shells, the other for non-interactive shells (e.g. plugin-host)
 COPY --chown=gitpod:gitpod nvm-lazy.sh /home/gitpod/.nvm/nvm-lazy.sh
-ENV PATH=/home/gitpod/.nvm/versions/node/v${NODE_VERSION}/bin:$PATH
--- a/tests/lang-node.yaml
+++ b/tests/lang-node.yaml
@@ -20,3 +20,13 @@
   entrypoint: [bash, -i, -c]
   assert:
   - status == 0
+- desc: it should yarn work
+  command: [yarn global add envinfo]
+  entrypoint: [bash, -i, -c]
+  assert:
+  - status == 0
+- desc: it should pnpm work
+  command: [pnpm i -g envinfo]
+  entrypoint: [bash, -i, -c]
+  assert:
+  - status == 0


chunks/lang-node/Dockerfile
```diff
@@ -9,12 +9,16 @@ USER gitpod
...
+ENV PNPM_HOME=/home/gitpod/.pnpm
+ENV PATH=/home/gitpod/.nvm/versions/node/v${NODE_VERSION}/bin:/home/gitpod/.yarn/bin:${PNPM_HOME}:$PATH
+
 RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | PROFILE=/dev/null bash \
     && bash -c ". .nvm/nvm.sh \
         && nvm install v${NODE_VERSION} \
         && nvm alias default v${NODE_VERSION} \
-        && npm install -g typescript yarn pnpm node-gyp" \
+        && npm install -g yarn \
+        && yarn global add typescript yarn pnpm node-gyp" \
```
gengjiawen, line 20: Please notice, this is intentional, for global packages survive across nvm versions.

gengjiawen:
@kylos101 Can you review this ?

chunks/lang-node/Dockerfile
axonasif: I'm not exactly familiar with js tooling. But I have a question, does it mean that if I switch to a different nodejs version with `nvm`, `typescript yarn pnpm node-gyp` executables will be using the old version interpreter to operate unless the user manually reinstall them in the changed version?

kylos101: Good point, @axonasif. 

@gengjiawen is this a standard practice, to install yarn globally, and then use yarn to install other packages globally? If yes, could we ask you to share a reference or two to help us better understand the trade-offs?

I don't see any references to `yarn global ...` [here](https://github.com/gitpod-io/gitpod), so it's not necessarily a standard that we're using "now".

Perhaps @gitpod-io/engineering-webapp can help here, as they work a bunch in node/typescript/yarn, too.

kylos101:
> @kylos101 Can you review this ?

hi @gengjiawen , can yo help us out [here](https://github.com/gitpod-io/workspace-images/pull/907#discussion_r955214907) to better understand? :pray:

chunks/lang-node/Dockerfile
gengjiawen: > But I have a question, does it mean that if I switch to a different nodejs version with nvm, typescript yarn pnpm node-gyp executables will be using the old version interpreter to operate unless the user manually reinstall them in the changed version?

Nope, those packages works as expected. They will be still work if you uninstall preinstalled nodejs.

> is this a standard practice, to install yarn globally

Yes. Same as pnpm and npm.

> then use yarn to install other packages globally? 

yeap. npm can do the same job. But nvm hijacked npm, so it only partly works.

> I don't see any references to yarn global ... [here](https://github.com/gitpod-io/gitpod), so it's not necessarily a standard that we're using "now".

Yeap. It's not needed if don't switch to another Node.js version or use yarn or pnpm install global packages.
But if you do, it's broken.

gengjiawen: Generally speaking, it's more robust and compatible with less surprise with current patch.

gengjiawen: @axonasif ping

kylos101: :wave: @gengjiawen I've had time to think about this change and scenario.

Please refer to [this matrix](https://gist.github.com/LayZeeDK/c822cc812f75bb07b7c55d07ba2719b3) for an example. You can see it relates compatible versions of Angular, Node, Typescript, and RxJS. So, users may wish to use nvm to switch between node/npm installations that are coupled to distinct Typescript installations (rather than have Typescript be installed globally across all node/npm versions). 

Given the above example, it would be more desirable to install Typescript with NPM, instead of yarn, so that they can use NVM to swap between node versions (and their related global dependencies, like Typescript).

If it's desired to use yarn to manage global dependencies (across all node versions which you can switch between node version using NVM), I recommend creating a custom base image, and manage that related flow from your own [custom Dockerfile](https://www.gitpod.io/docs/references/gitpod-dockerfile#custom-docker-image).

kylos101 (changes requested):
:wave: @gengjiawen may I ask you to review my thoughts (below) and share what you think? 

I believe we should not merge this PR given the example I cited below. Ultimately I think this will help Node.js usage remain flexible in Gitpod, while still granting folks the option to create their own custom image.

chunks/lang-node/Dockerfile
kylos101: Given :point_up:, unless we can reach an alternate solution, this PR will have to be closed w/o merge.

gengjiawen:
> Please refer to [this matrix](https://gist.github.com/LayZeeDK/c822cc812f75bb07b7c55d07ba2719b3) for an example. You can see it relates compatible versions of Angular, Node, Typescript, and RxJS. So, users may wish to use nvm to switch between node/npm installations that are coupled to distinct Typescript installations (rather than have Typescript be installed globally across all node/npm versions).

I can revert the typescript part if you like. But I do thinks put typescript in devDependencies is the right thing to do.

> If it's desired to use yarn to manage global dependencies

The patch to yarn is mainly to fix https://github.com/gitpod-io/workspace-images/issues/639. Yarn will keep packages global available and work in every conditions.

Also this patch also fix pnpm install global issues.

chunks/lang-node/Dockerfile
```diff
@@ -9,12 +9,15 @@ USER gitpod
...
```
Siddhant-K-code, line 20: I think we don't need a typescript here. But, yeah other packages should be installed globally.
```suggestion
        && yarn global add yarn pnpm node-gyp" \
```

chunks/lang-node/Dockerfile
geropl: @AlexTugarev you introduced nvm originally IIRC. Could you review? :thinking:

chunks/lang-node/Dockerfile
kylos101: `pnpm` can vary by node version too, https://pnpm.io/installation#compatibility, therefore, we should not install it with `yarn`, because then it won't change when we switch with `nvm`.

```suggestion
        && yarn global add yarn node-gyp" \
```

I would be fine with :point_up: , because `node-gyp` depends on the python version, and will not change with `nvm`.

gengjiawen: https://github.com/nodejs/Release Most tools drop support for un-maintained versions.

gengjiawen: If I move typescript install using npm. Is this PR can be approvoed ? @kylos101

gengjiawen: @kylos101 ping

gengjiawen: Friendly ping.

axonasif: Hey @gengjiawen, I understand that you're excited to get merged and I appropriate that you are 🙏 

@kylos101 is on vacation but will be back in a few days!

> can you help on this one?

I would love to but I'm not sure how can I help further here, I do not maintain the `workspace-images`, team-workspace does 🙏

gengjiawen: ping again.

gengjiawen: @sagor999 Anything you can help here ?

chunks/lang-node/Dockerfile
```diff
@@ -9,12 +9,15 @@ USER gitpod
 ENV TRIGGER_REBUILD=1
 ENV NODE_VERSION=${NODE_VERSION}
 
+ENV PNPM_HOME=/home/gitpod/.pnpm
+ENV PATH=/home/gitpod/.nvm/versions/node/v${NODE_VERSION}/bin:/home/gitpod/.yarn/bin:${PNPM_HOME}:$PATH
```
kylos101, line 13: Adding `PNPM_HOME` to the path is great, and fixes #913, and adding `/home/gitpod/.yarn/bin` fixes #639 , thank you @gengjiawen !

chunks/lang-node/Dockerfile
kylos101, line 20: @gengjiawen let's use the original code to install global dependencies for now:

```suggestion
        && npm install -g typescript yarn pnpm node-gyp" \
```

This way, users depending on `nvm` to switch between nodejs versions do not have broken workflows, where maybe they were expecting to have the ability to have different versions of tooling with different nodejs versions. In other words, global dependencies should be for a nodejs version, not the platform as a whole.

I recommend you maintain a custom image if you wish to use yarn to avoid having to reinstall global dependencies when changing nodejs versions with nvm. Also, if you'd like to socialize a change (for the future) to use yarn (like you did here), may I ask you to create an issue so we can socialize? This way, we can avoid delay this PR (which nicely fixes two issues already).

kylos101 (commented):
Hi @gengjiawen ,

You've done a wonderful job fixing two issues, I think we're close. 

I will apply my suggestion, because I'm trying to avoid a change to node base image, that could potentially cause unexpected behavior for users.

Then, I'll verify tests pass, and proceed with approval/merge. If you'd like to use yarn to install global dependencies, so that they survive nodejs changes via nvm, please consider a custom image, or create a separate issue so we can discuss.

chunks/lang-node/Dockerfile
gengjiawen: Yeah，I have put it pr description ，when merged should close related issues

kylos101: Indeed :) I was just sharing praise, but also highlighting the part of this PR that resolves the context issues.

kylos101 (approved):
:ship: it

kylos101:
Hi @gengjiawen once [this build](https://github.com/gitpod-io/workspace-images/actions/runs/3215815402) finishes, there will be new image tag with a timestamp close to "now" which you'll be able to use. 

Then, Tuesday, we'll tag the latest images as `latest`.

PR #922, opened 2022-09-03
fix: python extension settings write

axonasif:
## Related Issue(s)

Fixes:
Screenshot 2022-09-03 at 7 35 56 PM

## Release Notes

```release-note
```

### DIFF
--- a/chunks/lang-python/python_hook.bash
+++ b/chunks/lang-python/python_hook.bash
@@ -13,10 +13,12 @@ function pyenv_gitpod_init() {
 
 		if test ! -v GP_PYENV_INIT; then {
 
+			local vscode_machine_settings_file="/workspace/.vscode-remote/data/Machine/settings.json"
+			local ms_vscode_server_settings="$HOME/.vscode-server/data/Machine/settings.json"
+
 			function vscode::add_settings() {
 				# From https://github.com/axonasif/dotfiles/blob/main/src/utils/common.sh
 				local lockfile="/tmp/.vscs_add.lock"
-				local vscode_machine_settings_file="/workspace/.vscode-remote/data/Machine/settings.json"
 				local vscode_machine_settings_file="${SETTINGS_TARGET:-$vscode_machine_settings_file}"
 				trap 'rm -f $lockfile' ERR SIGINT RETURN
 				while test -e "$lockfile" && sleep 0.2; do {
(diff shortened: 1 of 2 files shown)


kylos101:
@axonasif can you add a brief description, how to test, and related release note? Also, it would be great if you could add a related test in `./tests` (although, that might be tough). :pray: :pleading_face: 

Note: I did `./build-combo.sh python-3.10` (to build and test the combo within a Gitpod workspace, it passed), and ran `docker run --rm -it localhost:5000/dazzle:python-3.10` and didn't notice any errors on start. How else might you test?

axonasif:
Hey @kylos101, yeah sorry haha 😅

It could be tested in such manner:
```bash
# The `$HOME/.bashrc/*-python` script relies on the existence of /workspace/.vscode-remote/data/Machine/settings.json as it expects a real-workspace, so we need GITPOD_REPO_ROOT to be there otherwise it won't make any change to the env.
docker run -v /workspace:/workspace -e GITPOD_REPO_ROOT -it <tag>
# Then see if any error message pops up
# To further verify, run:
cat /workspace/.vscode-remote/data/Machine/settings.json
# And observe that it's a valid json file with `python` settings added.
```
Or by pushing the image to dockerhub or somewhere else to use it via `.gitpod.yml`.

> Also, it would be great if you could add a related test in ./tests

I did now 🙏 , one of the reasons I'm hesitant to adding tests for such small changes is because it takes a long-while (at least ~5min as the minimum) as it has to rebuild (including the additional `base` every-time) the chunk(s) (including the additional `base` every-time) for any change inside `./tests`, which doesn't make any sense to me specially when you have to spend hours just to make a tiny new `test` work, it could've been a ~5 minute job in most cases 🥲 
And for the record, it took me around 41-minutes to make this new test work.

To cut the time I often time use `docker` to build during prototyping. And for `tests` I comment out every other `./tests/<something>.yml : -desc:` to only perform a specific test for a chunk but still it's very slow process.

jenting (approved):

axonasif:
Can we merge it now 🙏 ?

kylos101:
All set, @axonasif ! cc: @jenting 

You can build and test individual chunks (saving time) following [this example](https://github.com/gitpod-io/workspace-images/blob/main/CONTRIBUTING.md#build-specific-chunks). Does that help?

axonasif:
> @kylos101: You can build and test individual chunks (saving time) following [this example](https://github.com/gitpod-io/workspace-images/blob/main/CONTRIBUTING.md#build-specific-chunks). Does that help?

I do that but only when I have to run the `/tests`, otherwise I test a chunk `Dockerfile` by manually using `docker` CLI as thats a faster process, I wish dazzle used caching instead of rebuilding things each time and rebuilding without any change inside `chunks/` dir when you modify `tests/`.

PR #928, opened 2022-09-21
refactor: python_hook.bash

axonasif:
## Description

This time it will handle errors properly and will not log anything on the attached pty but write to `/tmp/.vscode::add_settings.log` file if any error occurs. Will always return true on the parent shell process.

## Related Issue(s)

~~Reverts the change that inconsistently causes [this](https://gitpod.slack.com/archives/C01KGM9BH54/p1663747428135699)
It was mainly due to the parent directory not being present before the file creation. But since https://github.com/gitpod-io/gitpod/issues/13029 is being worked out, so we no longer need to fix it here or handle this.~~ In the future we can get rid of this function and use a native built-in gitpod method once we have.

## How to test

```bash
docker build --build-arg base=gitpod/workspace-base --build-arg PYTHON_VERSION=3.8.13 -t python chunks/lang-python

# The `$HOME/.bashrc/*-python` script relies on the existence of $GITPOD_REPO_ROOT as it expects a real-workspace, so we need GITPOD_REPO_ROOT to be there otherwise it won't make any change to the env.

# Test without an existing settings.json
rm -f /workspace/.vscode-remote/data/Machine/settings.json && docker run -v /workspace:/workspace -e GITPOD_REPO_ROOT -it python

# Test with an existing settings.json
printf '{ "%s": "%s" }\n' one value > /workspace/.vscode-remote/data/Machine/settings.json && docker run -v /workspace:/workspace -e GITPOD_REPO_ROOT -it python

# Test with a settings.json containing syntax errors
printf '{ "%s" "%s" \n' one value > /workspace/.vscode-remote/data/Machine/settings.json && docker run -v /workspace:/workspace -e GITPOD_REPO_ROOT -it python

# To further verify, run:
cat /workspace/.vscode-remote/data/Machine/settings.json ~/.vscode-server/data/Machine/settings.json
# And observe that it's a valid json file with `python` settings added.

# Check log file
cat "/tmp/.vscode::add_settings"
```

## Release Notes

```release-note
```

### DIFF
--- a/chunks/lang-python/python_hook.bash
+++ b/chunks/lang-python/python_hook.bash
@@ -13,53 +13,47 @@ function pyenv_gitpod_init() {
 
 		if test ! -v GP_PYENV_INIT; then {
 
-			local vscode_machine_settings_file="/workspace/.vscode-remote/data/Machine/settings.json"
-			local ms_vscode_server_settings="$HOME/.vscode-server/data/Machine/settings.json"
-
-			function vscode::add_settings() {
+			function vscode::add_settings() (
 				# From https://github.com/axonasif/dotfiles/blob/main/src/utils/common.sh
+				exec >>"/tmp/.${FUNCNAME[0]}.log" && exec 2>&1
+				set -eu
 				local lockfile="/tmp/.vscs_add.lock"
-				local vscode_machine_settings_file="${SETTINGS_TARGET:-$vscode_machine_settings_file}"
-				trap 'rm -f $lockfile' ERR SIGINT RETURN
+				trap 'rm -f $lockfile $tmp_file || :;:' ERR SIGINT RETURN
 				while test -e "$lockfile" && sleep 0.2; do {
 					continue
 				}; done
 				touch "$lockfile"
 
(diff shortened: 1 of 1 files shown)


jeanp413:
@axonasif do you know what I'm missing? I get this error when running the second command
```
Unable to find image 'localhost:5000/dazzle:lang-python-3.8' locally
docker: Error response from daemon: manifest for localhost:5000/dazzle:lang-python-3.8 not found: manifest unknown: manifest unknown.
```

axonasif:
Hey @jeanp413, please check the description again, I updated it :pray:

jeanp413 (approved):

kylos101 (approved):
Thank you @axonasif ! :hugs:

kylos101:
@axonasif @Siddhant-K-code this is live now on Docker Hub (just pushed from main).

kylos101:
@jeanp413 @akosyakov FYI, this shipped just now :wave:

Siddhant-K-code:
Thanks @kylos101 🙌🏼

PR #814, opened 2022-04-22
Update Ubuntu from focal (20.04) to jammy (22.04)

tnir:
Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>

## Description
Updates Ubuntu from focal (20.04) to jammy (22.04).

## Related Issue(s)
Fixes #810

## How to test
TBD

## Release Notes
```release-note
!Bump ubuntu from 20.04 to 22.04
```

## Documentation
Does this PR require updates to the documentation at www.gitpod.io/docs?
* Yes
  * 1. Please create a docs issue: https://github.com/gitpod-io/website/issues/new?labels=documentation&template=DOCS-NEW-FEATURE.yml&title=%5BDocs+-+New+Feature%5D%3A+%3Cyour+feature+name+here%3E
  * 2. Paste the link to the docs issue below this comment

Issue #810, opened 2022-04-22
jammy (Ubuntu 22.04 LTS)- based images


tnir:
jammy (Ubuntu 22.04 LTS) ~~is coming~~ is just released.

What is the strategy for jammy-based workspace images?

### Resources
- https://lists.ubuntu.com/archives/ubuntu-announce/2022-April/000279.html
- The latest update: #197
   - also #196

princerachit:
@gitpod-io/engineering-workspace do we have any hard dependency on ubuntu version that might break with this change. Things like existing work so far with K3s in gitpod / tailscale etc?

We can curate a list and then make sure that we test it before releasing related change to the public. Our existing tests are not enough IMO to just merge a base image verion upgrade change.

tnir:
Prebuild system and underlying Kubernetes should be updated. Details were described in the following comment:

- https://github.com/gitpod-io/workspace-images/pull/814#issuecomment-1108091553

Versioning with Docker image tagging would be helpful to maintain the stack, like `ubuntu:focal` and `ubuntu:jammy` (now equivalent to `ubuntu:latest`). Users should use `gitpod/workspace-full:focal` and `gitpod/workspace-full:jammy` instead of `gitpod/workspace-full`.

Furisto:
> @gitpod-io/engineering-workspace do we have any hard dependency on ubuntu version that might break with this change. Things like existing work so far with K3s in gitpod / tailscale etc?

Not that I am aware of.

ianhellstrom:
It would be great if users could start testing with Jammy. Any updates here?

kylos101:
Hi @tnir @ianhellstrom and @david-bakin,

It just so happens that [we're updating our documentation to help users create custom images which do not use Gitpod base images](https://github.com/gitpod-io/website/issues/2850). For example, if your workflow cannot use a Gitpod base image, the docs will show customizations that are needed to have a successful image build, as well as a nice development experience in Gitpod. This may also be a good short term solution for you if your workflow requires Jammy.

Also, we plan to [publish a version of `workspace-full` that supports Jammy](https://github.com/gitpod-io/workspace-images/issues/949). It won't be the default base image, but, it'll be available as something like`workspace-full:jammy.rc0`, etc. The reason for this is [so that we can use it](https://github.com/gitpod-io/gitpod/issues/13774) (to test internally), but also give the community something to test with, too. Once the release candidate is available on Docker Hub, we'll let y'all know.

[...]

PR #932, opened 2022-09-25
Update Ubuntu from focal (20.04) to jammy (22.04)

armanbilge:
Attempting to revive https://github.com/gitpod-io/workspace-images/pull/814 by @tnir.


david-bakin:
:-1: :-1: Is an upgrade of Ubuntu between major versions really so trivial it can be done with a PR and no testing in deployment by customers?  I don't know, which is why I'm asking.  It's not simply the Linux kernel - it's also all the mainstream repositories holding binaries and _development headers and libraries_.  What happened the _last_ time Gitpod upgraded Ubuntu versions?  (Or has it ever?)

Couldn't it mean, for example, that lots of customers will now have .gitpod.yml start tasks and custom Dockerfiles that won't build because the upstream apt repos are different?  Shouldn't there at a minimum be a "legacy" workspace-base maintained, with perhaps a different name such as workspace-base-20.04, so that customers can at least try to recover their old build environment?  Perhaps the entire _stack_ of different workspace images (or at a minimum, a `workspace-full-20.04` be built and kept until the impact is known?  Maybe all it takes is to _tag_ the last known good 20.04 images of each variety?   I ask again: shouldn't there be testing, in deployment, by actual users???  

Perhaps a _better_ approach would be to build the entire stack of images `-base`, `-cpp`, `-python`, `-full` etc. with different names (e.g., `workspace-full-22.04`) and encourage customers to _try them out_ with their own workspaces, see what happens?

(Is there a design document - that I'm not privy too - that goes into the risks/benefits?  Has anyone checked to see what competitor cloud development services do to upgrade the base system - or discovered that they know there are no problems?)

Seems like there's potential for making the customer experience _very_ destabilizing!  The base PR #814 has _no_ discussion of this itself but defers to #810 which only discusses it from Gitpod's internal point-of-view, not anything the _customer_ might experience.  Yet note that even #814 has [this comment](https://github.com/gitpod-io/workspace-images/pull/814#issuecomment-1108091553) pointing to at least one problem in someone else's dockerfile when upgrading to 22.04 so **_who knows what will happen to actual customers with their workspaces???_**

> Thanks. I did not intend to merge this within days since I already know the problem at least on Gitpod cloud (gitpod.io).
>
> cf. https://github.com/tianon/docker-brew-ubuntu-core/issues/236

(Gitpod stability being already a concern of customers means this upgrade should _not be taken lightly_.)

armanbilge:
> Is an upgrade of Ubuntu between major versions really so trivial it can be done with a PR and no testing in deployment by customers?

I agree, it's not trivial.

It's a shame that GitPod didn't make an image available earlier so users could begin trying it. For example GitHub actions made 22.04 available for beta testing originally in April before making it Generally Available in August. It will replace their default CI runner in the near future, at which point 20.04 will have to be opt-in.

But Ubuntu 22.04 is here, it's LTS, and soon it will be default in GitHub CI. That means (at least for me) it will be a problem if I cannot easily use it when I open any repository on GitHub with GitPod (i.e., not requiring custom configuration).

ianhellstrom:
I’d be happy to try it out, if I had the opportunity.

kylos101:
Closing in favor of https://github.com/gitpod-io/workspace-images/pull/1017

PR #943, opened 2022-10-07
feat: add linuxbrew completion support for bash

shyim:
Code is from https://docs.brew.sh/Shell-Completion#configuring-completions-in-bash


jenting:
CI failed. Could you please check it?
Thank you.

shyim:
@jenting Is it fine if I disable that shellcheck's? (like in python) I would not like to modify the original code from Homebrew :sweat_smile:

chunks/tool-brew/completion-hook.bash
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+if type brew &>/dev/null
+then
+  HOMEBREW_PREFIX="$(brew --prefix)"
+  if [[ -r "${HOMEBREW_PREFIX}/etc/profile.d/bash_completion.sh" ]]
+  then
+    source "${HOMEBREW_PREFIX}/etc/profile.d/bash_completion.sh"
```
jenting, line 8: You could add
```shell
# shellcheck disable=SC1091
```

chunks/tool-brew/completion-hook.bash
```diff
@@ -0,0 +1,15 @@
...
+  else
+    for COMPLETION in "${HOMEBREW_PREFIX}/etc/bash_completion.d/"*
+    do
+      [[ -r "${COMPLETION}" ]] && source "${COMPLETION}"
```
jenting, line 12: You could add
```shell
# shellcheck disable=SC1090
```

jenting (commented):
Reference to https://github.com/gitpod-io/workspace-images/blob/main/chunks/lang-node/nvm-lazy.sh#L21

chunks/tool-brew/completion-hook.bash
```diff
@@ -0,0 +1,15 @@
...
+    done
+  fi
+fi
```
axonasif, line 15: There is no need to implement custom logic.
Use `homebrew`'s shellenv instead.

```suggestion
#!/usr/bin/env bash

homebrew_bin="/home/linuxbrew/bin/brew"

if test -x "$homebrew_bin"; then {
    eval "$("$homebrew_bin" shellenv)"
} fi

unset homebrew_bin
```

axonasif (changes requested):
It's better to use the builtin `shellenv` of `brew` that loads completions and even more instead of implementing it ourselves. Also, please rename `completion-hook.bash` to something like `homebrew.bash` and update the path on your `Dockerfile` afterwards.

shyim:
@axonasif `/home/linuxbrew/.linuxbrew/bin/brew shellenv` does nothing.

It prints nothing out. Are you sure? :D

shyim:
AHHH you are talking about completion support for homebrew self.

I am talking about installed Software using homebrew 😂

axonasif:
> AHHH you are talking about completion support for homebrew self.
> 
> I am talking about installed Software using homebrew 😂

I'm not, please check yourself.

axonasif:
> @axonasif `/home/linuxbrew/.linuxbrew/bin/brew shellenv` does nothing.
> 
> It prints nothing out. Are you sure? :D

It's not supposed to print out anything if the necessary variables are already set.

shyim:
Never mind it works already without my changes. I was drunken or so. I have no idea.

axonasif:
> Never mind it works already without my changes. I was drunken or so. I have no idea.
> 
> Sorry 😢

Don't be sorry 😅 . Your changes are [valid](https://docs.brew.sh/Shell-Completion#configuring-completions-in-bash).

However, we're already using `bash-completion` package that is loaded beforehand, which is why it works without the change. Loading it from homebrew again might not be useful for most well-known packages.

shyim:
yea i guess.

The real issue was that I just have to reopen the bash

```bash
brew install FriendsOfShopware/tap/shopware-cli
shopware-cli <tab> does not work
```

reopen a terminal

```
shopware-cli <tab> works now
```

PR #945, opened 2022-10-09
feat: make user aware of the current workspace-image

axonasif:
## Description

- Show the current workspace-image info only when the user is using an official gitpod image. (by parsing `.gitpod.yml` and the dockerfile if-exists)
- Include a link to the source code of the current workspace-image that also contains a dedicated `README.md`
- Once every session whenever someone creates a new terminal manually. After first execution, the script removes itself.
- Will avoid task terminals, because showing there would not be very visible most of the time.
- User can completely disable this by adding `GITPOD_DISABLE_WORKSPACE_IMAGE_INFO=true` at https://gitpod.io/variables with `*/*` as the scope.

|base|full|
|---|---|
|Screenshot 2022-10-09 at 3 59 42 PM|Screenshot 2022-10-09 at 4 04 05 PM|
|full-vnc|other|
|Screenshot 2022-10-09 at 4 01 14 PM|Screenshot 2022-10-09 at 4 02 09 PM|

## Related Issue(s)

Relates with https://github.com/gitpod-io/gitpod/issues/10683

Next steps:
- [ ] Add README to each `chunk/*/` dir explaining what this chunk contains and if there's something special to be aware of.
- [ ] Add README to `base/` dir to explain how the base image sets up the required things for a minimal but functional docker image. In case someone wants to create their own minimal base image of a different distro (i.e Arch or Fedora)
- [ ] Add workspace-images changelog on these READMEs
- [ ] We could repurpose the READMEs for https://github.com/gitpod-io/dazzle/pull/56 when it's used here.

So that the user can click the scoped link from their terminal to read these READMEs

## How to test

- Open this PR on Gitpod.
- Run:
```bash
docker build -t base base/ && docker run -v /workspace:/workspace -e GITPOD_REPO_ROOT -e GITPOD_GIT_USER_NAME -it base
```
- Change the `image` values on `.gitpod.yml` and rerun the above command to get different info. Or you can remove `.gitpod.yml` and it will assume the default image, which is `workspace-full`. You may change the image name on `.gitpod.Dockerfile` as well.

## Release Notes

```release-note
Current workspace-image info is shown once every session on the terminal
```

### DIFF
--- a/base/bashrc.d/900-ws-info
+++ b/base/bashrc.d/900-ws-info
@@ -1,0 +1,137 @@
+(
+	set -eu
+	function remove_self() {
+		# Remove the hook file, because we no longer need it to be loaded.
+		if [[ "${BASH_SOURCE[0]}" =~ \.bashrc.d/[0-9]+- ]]; then {
+			rm -f "${BASH_SOURCE[0]}"
+		} fi
+	}
+	lockfile=/tmp/.wsinfo_lck
+	trap '{ remove_self; rm -f "$lockfile"; } || :' EXIT ERR
+
+	# Return early if user opted out
+	if test "${GITPOD_DISABLE_WORKSPACE_IMAGE_INFO:-false}" == true; then {
+		exit 0 # This is a subshell
+	} fi
+
+	if test ! -e "$lockfile"; then {
+		remove_self
+		printf '' > "$lockfile"
+
+		# The supervisor creates the task terminals, supervisor calls BASH from `/bin/bash` instead of the realpath `/usr/bin/bash`
+		# So we will only print the info once, on a non-task terminal that the user created manually
+		if ! [ "$BASH" == /bin/bash ] || ! [ "$PPID" == "$(pgrep -f "supervisor run" | head -n1)" ]; then {
+
+			# Get image source
+			gitpod_yml_path="$GITPOD_REPO_ROOT/.gitpod.yml"
+			if test ! -e "$gitpod_yml_path"; then {
+				: "workspace-full"
+			} elif res="$(awk '/image:/,/file:/' "$gitpod_yml_path" | awk '{ if (NR==2) { print $2; exit } }')" && test -n "${res:-}"; then {
+				if res="$(awk '{ if ($1 ~ /^FROM/) { print  $2; exit } }' "$GITPOD_REPO_ROOT/$res")" && test -n "${res:-}"; then {
+					: "${res#*/}"
+				} else {
+					: "workspace-full"
+				} fi
+			} elif res="$(sed -n 's/^image: //p' "$gitpod_yml_path")" && test -n "${res:-}"; then {
+				: "${res#*/}"
+			} else {
+				: "workspace-full"
+			} fi
+			image_source="$_"
+
+			wsimgs_repo_root="https://github.com/gitpod-io/workspace-images/tree/main"
+			username="${GITPOD_GIT_USER_NAME:-"gitpodder"}"
+			published_wsimgs=(
+				"workspace-base"
+				"workspace-c"
+				"workspace-cojure"
+				"workspace-elixir"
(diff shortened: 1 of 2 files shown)


kylos101:
:wave: @axonasif thanks for sharing, we really appreciate your effort with `workspace-images`. :smile_cat: 

Could I ask some favors?

1. Move this PR back to draft
2. Help us [breakdown](https://www.notion.so/gitpod/Development-Process-2-0-6681854173ab4d2f92880f9f3d85cab5#4cd86b05a62e4d05a6968c47d402528a) the related [epic](https://github.com/gitpod-io/gitpod/issues/10683), by creating and sharing a [RFC to socialize what you've done in this PR as a 3rd option, sharing workspace image content in the terminal](https://www.notion.so/gitpod/How-we-work-0f9159f282d24f9d9776372e0c7dbab4#f2a49a65e71144c884f0a716d3b23092). For example, you can link to this PR in the RFC, but also outline why you think it's appropriate, but also what other options you considered. 

Why?
1. For epics, it's important to socialize and agree on options and next steps, prior to authoring and sharing related PRs with teams.
2. I'd like to avoid adding script to `workspace-images`, if possible, because it can create maintenance needs and complexity, potentially even conflicting with the IDE experience. 

I like option 2, for example, because as a :skateboard: we could improve documentation and readme content, to help improve the user experience. A benefit of this would be that (1) we're not adding complexity to `workspace-images` and (2) not risking IDE compatibility issues (like we recently had with Python and VSCode).

base/bashrc.d/900-ws-info
```diff
@@ -0,0 +1,137 @@
...
+	if test "${GITPOD_DISABLE_WORKSPACE_IMAGE_INFO:-false}" == true; then {
+		exit 0 # This is a subshell
+	} fi
+
+	if test ! -e "$lockfile"; then {
+		remove_self
+		printf '' > "$lockfile"
+
+		# The supervisor creates the task terminals, supervisor calls BASH from `/bin/bash` instead of the realpath `/usr/bin/bash`
+		# So we will only print the info once, on a non-task terminal that the user created manually
```
kylos101, line 22: @axonasif as an alternative, it would be nice if this output were visible in a separate (exclusive) task terminal, avoiding user opened terminals altogether. In other words, then, they could always "reference it", by simplying looking at it, without having to ever opt-out.

Another thought is, we persist this output somewhere in `/workspace`, and then link to it from the Gitpod menu "in workspace". In other words, never present this content in a Terminal.

:point_up: are example thoughts that would be best to socialize as feedback in a RFC, as an example. :pray:

akosyakov:
Would it pop up in each terminal?

axonasif:
> Would it pop up in each terminal?

Hey @akosyakov, it will be printed once for a user created terminal.
More info on this [RFC](https://www.notion.so/gitpod/Show-workspace-image-info-on-the-terminal-595cc898c1e947caa78d2b522a6aa16c) :pray:

akosyakov:
> Hey @akosyakov, it will be printed once for a user created terminal.
More info on this [RFC](https://www.notion.so/gitpod/Show-workspace-image-info-on-the-terminal-595cc898c1e947caa78d2b522a6aa16c) 🙏

Not for supervisor task terminals right?

base/bashrc.d/900-ws-info
akosyakov: I think it is good idea to have one task terminal which shows such info. I also like idea of doing it in supervisor only for regular workspaces.

axonasif:
> > Hey @akosyakov, it will be printed once for a user created terminal.
> > More info on this [RFC](https://www.notion.so/gitpod/Show-workspace-image-info-on-the-terminal-595cc898c1e947caa78d2b522a6aa16c) 🙏
> 
> Not for supervisor task terminals right?

Right. It won't show on them.

PR #958, opened 2022-10-16
Proposal: feat: Add docs explaining each chunk and include hints

axonasif:
## Description

This was one of the **Next Steps** on https://github.com/gitpod-io/workspace-images/pull/945
I'm not sure if we will move ahead with the proposed solution on this [RFC](https://www.notion.so/gitpod/Show-workspace-image-info-on-the-terminal-595cc898c1e947caa78d2b522a6aa16c), which was https://github.com/gitpod-io/workspace-images/pull/945.

Since it was one of the next steps of https://github.com/gitpod-io/workspace-images/pull/945 and everyone suggested for the improvements regarding documentation on RFC, so that's why bringing this up in advance.

Preview can be seen [here](https://github.com/gitpod-io/workspace-images/tree/axonasif/add_docs/base#workspace-base) and [here](https://github.com/gitpod-io/workspace-images/tree/axonasif/add_docs/chunks#workspace-full).

cc @kylos101 @atduarte

## TODOs
These chunks are left to be documented:
- [x] lang-ruby
- [x] tool-{brew, docker, tailscale, dotnet, mongodb, nginx, postgresql}

## Next steps
- [ ] Update https://www.gitpod.io/docs/configure/workspaces/workspace-image pointing out about the READMEs on this repo.
- [x] Update the main README.md on this repo to indicate that ☝️ as well.

## Release Notes

```release-note
```

## Documentation

Yes. See **Next Steps**.

### DIFF
--- a/chunks/lang-ruby/README.md
+++ b/chunks/lang-ruby/README.md
@@ -16,9 +16,9 @@ Lastly, [see it in action!](https://www.gitpod.io/docs/introduction/learn-gitpod
 
 We provide multiple variants of the `gitpod/workspace-ruby-*` docker image as per [chunk.yaml](./chunk.yaml). Each containing a different version of Ruby:
 
-- `gitpod/workspace-ruby-2.7` (This is ruby 2.7.6)
-- `gitpod/workspace-ruby-3.0` (This is ruby 3.0.4)
-- `gitpod/workspace-ruby-3.1` (This is ruby 3.1.2)
+- `gitpod/workspace-ruby-2.7` (This being ruby 2.7.x)
+- `gitpod/workspace-ruby-3.0` (This being ruby 3.0.x)
+- `gitpod/workspace-ruby-3.1` (This being ruby 3.1.x)
 
 You could use these on your `.gitpod.yml` as well.
 
--- a/chunks/lang-ruby/README.md
+++ b/chunks/lang-ruby/README.md
@@ -1,0 +1,52 @@
+# workspace-ruby-*
+
+## How to use for a Gitpod workspace
+
+Put the following line in your `.gitpod.yml`:
+
+If you do not have a `.gitpod.yml`, run `gp init` on your terminal to create one.
+
+```yaml
+image: gitpod/workspace-ruby-3.1
+```
+
+Lastly, [see it in action!](https://www.gitpod.io/docs/introduction/learn-gitpod/gitpod-yaml#see-it-in-action)
+
+### Extra docker image variants
+
+We provide multiple variants of the `gitpod/workspace-ruby-*` docker image as per [chunk.yaml](./chunk.yaml). Each containing a different version of Ruby:
+
+- `gitpod/workspace-ruby-2.7` (This is ruby 2.7.6)
+- `gitpod/workspace-ruby-3.0` (This is ruby 3.0.4)
+- `gitpod/workspace-ruby-3.1` (This is ruby 3.1.2)
+
+You could use these on your `.gitpod.yml` as well.
+
+## Details
+
+This is mainly optimized for [Ruby](https://www.ruby-lang.org/en/), and is based on [workspace-base](../../base/)
+
+The following are installed for you, you can extend using a [custom-docker-image](https://www.gitpod.io/docs/configure/workspaces/workspace-image#configure-a-custom-dockerfile) if you wish.
+
+- We use `rvm`.
+    - With `rvm` a complete Ruby runtime is installed in the user-land, the version [depends on the image you're using](#extra-docker-image-variants).
+- Some extra packages are installed on the user-land installation with `gem`:
+    - bundler, solargraph
+
+For more details, check the [Dockerfile](./Dockerfile)
+
+### To be aware of
+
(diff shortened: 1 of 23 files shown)


chunks/lang-ruby/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Lastly, [see it in action!](https://www.gitpod.io/docs/introduction/learn-gitpod/gitpod-yaml#see-it-in-action)
+
+### Extra docker image variants
+
+We provide multiple variants of the `gitpod/workspace-ruby-*` docker image as per [chunk.yaml](./chunk.yaml). Each containing a different version of Ruby:
+
+- `gitpod/workspace-ruby-2.7` (This is ruby 2.7.6)
+- `gitpod/workspace-ruby-3.0` (This is ruby 3.0.4)
+- `gitpod/workspace-ruby-3.1` (This is ruby 3.1.2)
```
atduarte, line 21: This is will rapidly get out of date. Could we change to the following:
```suggestion
- `gitpod/workspace-ruby-2.7` (This being ruby 2.7.x)
- `gitpod/workspace-ruby-3.0` (This being ruby 3.0.x)
- `gitpod/workspace-ruby-3.1` (This being ruby 3.1.x)
```

It applies also to other images of different languages.

atduarte (commented):
I like this a lot, @axonasif! Thanks so much for your making this happen. Really appreciate your initiative 🧡

The text looks on point. Although I'm not sure they belong inside the chunks, as an image can be (and are) created from multiple different chunks. What if we moved all this documentation to the root README.md making it collapsible? 🤔

axonasif:
Thank you for stepping in to review @atduarte 🙌 

> Although I'm not sure they belong inside the chunks, as an image can be (and are) created from multiple different chunks.

Yeah, which chunks are used to build a whole-image is mentioned in each of these READMEs.

> What if we moved all this documentation to the root README.md making it collapsible?

I think it's less overwhelming when you can navigate _in chunks ( 😝 )_ with context of the source-code and only get the info you're looking for. Most of the relevant things are cross linked.

chunks/lang-ruby/README.md
axonasif: Thanks for pointing this out, will do.

axonasif:
Made the changes regarding [this](https://github.com/gitpod-io/workspace-images/pull/958#discussion_r998061501).
Please let me know what else needs to be done @atduarte 😄

Siddhant-K-code:
It's a good idea, but I do not think we can scale it up (with a lot of manual work). That was the intention from this PR https://github.com/gitpod-io/dazzle/pull/56 to automate this work with every new release.

atduarte:
@Siddhant-K-code I think it's ok and potentially better to do it manually, discuss, iterate, and only then automate. As automation is harder to iterate on.

@axonasif I _really_ would prefer to avoid mixing the concepts of images and chunks. They are very different and it will create issues. Examples of problems that would exist right away:
1. dep-cacert-update, tool-docker and tool-tailscale do not result in a specific image
2. tool-chrome results in multiple images

Said that I would propose we either:
1. add a list of images with collapsed description in the README
2. add a separate folder called docs/
3. make this documentation part of the website docs (preferred)

In all these cases, we would point to the respective base image and chunks that make it.

axonasif:
> I really would prefer to avoid mixing the concepts of images and chunks. They are very different and it will create issues

@atduarte fair point, I didn't think about that 😅 

> tool-docker and tool-tailscale do not result in a specific image
> tool-chrome results in multiple images

Yeah, currently it's only mentioned that no individual public image is produced from these[^1][^2].

> make this documentation part of the website docs (preferred)

I would love to do that too, but need guidance on how to surface this on our docs. Additionally, I would also like to keep it very related with the source code even if we host the docs externally. [vscode-dev-containers](https://github.com/microsoft/vscode-dev-containers/tree/main/containers) also follow a similar pattern I think.

[^1]: https://github.com/gitpod-io/workspace-images/tree/axonasif/add_docs/chunks/tool-tailscale#details
[^2]: https://github.com/gitpod-io/workspace-images/tree/axonasif/add_docs/chunks/tool-docker#details

atduarte:
> I would love to do that too, but need guidance on how to surface this on our docs 🙏 

@axonasif Do you have any ideas? Thought about a reference page for "Workspace Images" and linking that from https://www.gitpod.io/docs/configure/workspaces/workspace-image#use-a-public-docker-image. 

Calling in the unofficial Docsking, @loujaybee 👑 😁 

> Additionally, I would also like to keep it very related with the source code

Same, but that is not a new problem, right? 🤔 It also exists with code changes to gitpod-io/gitpod. So we could figure out how we resolve it later.

axonasif:
This PR was moved back to draft based on discussion from [here](https://www.notion.so/gitpod/Discussion-about-documenting-workspace-images-ffcb0668494841c4bf0bef0dc9dba8a3) (internal)

We decided to create a brief intro page for workspace-full first and then move ahead with some changes.

axonasif:
@filiptronicek sure, added this to my todo list to follow up later

PR #967, opened 2022-10-29
feat: update composer to 2.4.4

shyim:
- Simplified install of composer by just copying it from the official docker image
- Added PHP-FPM to speed up [Symfony-CLI](https://symfony.com/doc/current/setup/symfony_server.html)

### DIFF
--- a/chunks/tool-nginx/Dockerfile
+++ b/chunks/tool-nginx/Dockerfile
@@ -4,7 +4,10 @@ FROM ${base}
 USER root
 
 # Dazzle does not rebuild a layer until one of its lines are changed. Increase this counter to rebuild this layer.
-ENV TRIGGER_REBUILD=1
+ENV TRIGGER_REBUILD=2
+
+# Install composer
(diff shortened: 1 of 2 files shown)


chunks/tool-nginx/Dockerfile
```diff
@@ -4,7 +4,10 @@ FROM ${base}
 USER root
 
 # Dazzle does not rebuild a layer until one of its lines are changed. Increase this counter to rebuild this layer.
-ENV TRIGGER_REBUILD=1
+ENV TRIGGER_REBUILD=2
+
+# Install composer
+COPY --from=composer:2.4.4 /usr/bin/composer /usr/bin/composer
```
shyim, line 10: We could also pin it to `:2` so it's getting automatically updated and does not contain any breaking change

sagor999: I like the idea of pinning to :2 
Would need to update the test below as well to make sure it does not break when new version comes out. 
would you mind making that change? 🙏

sagor999 (approved):
🧡

PR #977, opened 2022-11-09
Updated dotnet version from `dotnet 6` to `dotnet 7`

Mercurial:


sagor999:
@Mercurial would this break anyone's workspace when their image gets upgraded? :thinking: 
Is version 7 backwards compatible with 6?

Mercurial:
Yes it's backwards compatible it can run and compile dotnet 6 code at the very least

Mercurial:
> @Mercurial it seems like some of the tests are failing. Can you please take a look? Thank you!

Found the issue, it seems like the test was hard-coded to look for the `Microsoft.AspNetCore.App` version `6` string in `dotnet --info` command which I don't have context why is that since its not a good measure of compatibility. 

I have pushed a commit to check for the proper version but again I dont have any context if this test is good or bad.

sagor999 (approved):

PR #983, opened 2022-11-25
feat(base): autoload bash env for other shells

axonasif:
## Description

This adds startup scripts specific to fish and zsh to autoload the `bash` environment variables within their sessions without altering the user configuration (i.e. dotfiles)

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10105

## How to test

Screenshot 2022-11-25 at 3 36 29 PM

Open this PR on Gitpod.

Build and run:
```bash
cd base
docker build -t base . && docker run -it base
```
We will land in a `bash` shell. Now we can create a script inside `~/.bashrc.d` from the bash session and then switch to another shell to test if it picked it up.

```bash
echo 'export TESTVAR=hello' > ~/.bashrc.d/testing

# Switch to fish
fish

# Check the variable
echo $TESTVAR

# Exit
exit

# Switch to zsh
zsh

# Check
echo $TESTVAR

# Exit
exit

# We'll be back on `bash` (however, it didn't pick up `~/.bashrc.d/testing`, because it was created during the session)
echo $TESTVAR # will be empty
# let's check in `env` as well
env | grep TESTVAR # will be empty

# Now let's reload bash session
exec bash -l

# Now it should be available on bash as well
env | grep TESTVAR
```

## Release Notes

```release-note
```

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -59,6 +59,13 @@ COPY --chown=gitpod:gitpod default.gitconfig /home/gitpod/.gitconfig
 # configure git-lfs
 RUN git lfs install --system --skip-repo
 
+# Autoload .bashrc.d for other SHELLs (fish and zsh)
+RUN mkdir -m 755 -p /usr/share/fish/vendor_conf.d /etc/zsh \
+    && printf 'source %s\n' /etc/zsh/autoload_bash.zsh >> /etc/zsh/zshrc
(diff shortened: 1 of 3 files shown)


axonasif:
This could be implemented via `supervisor` as well, we just need to write two files.

axonasif:
Converted to draft.
Context [here](https://gitpod.slack.com/archives/C01KGM9BH54/p1669363120784239).

PR #1017, opened 2023-01-31
Change base image to Ubuntu Jammy

kylos101:
## Description

Breaking changes
1. Update maintained images[[1](https://github.com/gitpod-io/workspace-images/#-images-well-maintain)][[2](https://github.com/gitpod-io/workspace-images/#-specific-images)] at https://hub.docker.com/u/gitpod from Focal to Jammy
2. [Deprecate and do not update `gitpod/workspace-ruby-2`; Ruby 2.7 is EOL March 31, 2023](https://endoflife.date/ruby).
3. Upgrade `gitpod/workspace-mongodb` to mongo v6. v5 doesn't support OpenSSL 3, [OpenSSL 3 comes out of the box with Jammy](https://packages.ubuntu.com/source/jammy/openssl).
4. Update `gitpod/workspace-node-lts` from Node 16 to Node 18, and also in `gitpod/workspace-full`
5. Update `gitpod/workspace-node` from Node 18 to Node 19
6. Install Python 3.9 in `gitpod/workspace-yugabytedb*`, changing its `base` ref from `full`. Refer to its [dockerfile](https://github.com/gitpod-io/workspace-images/blob/kylos101/jam/chunks/tool-yugabytedb/Dockerfile#L27) for background

Fixes
1. Replace Go 1.18 (which was failing the build on `dazzle build` and not part of any dazzle combination / workspace image), with Go 1.20. 
   * Continue using 1.19.5 in the go combination (so there is no change to users) for `gitpod/workspace-full` or `gitpod/workspace-go`.
   * We should switch to 1.20 as part of the next release, @atduarte and @loujaybee for :eyes: 
2. Bump Rust to 1.67.1
3. Bump Ruby patch versions for 3.0, 3.1, and 3.2
4. Override OpenSSL for Ruby 3.0 to support Jammy
5. Bump Ruby in `gitpod/workspace-full` from Ruby 3.1 to 3.2
6. Bump Python in `gitpod/workspace-full` from Python 3.8 to 3.11
7. Bump nvm version from 0.39.0 to 0.39.3
8. Bump buildkit and dazzle dependencies for CI

New features
1. Use asdf version manager in Elixir combination to more easily manage erlang and elixir

## Related Issue(s)

Fixes: #810, #1016, #1029 and #1012

## How to test

- Our tests pass in this build.
- After this merges to main
[...]

PR #1041, opened 2023-02-16
update Go to `1.20.6`

0xmohit:
## Description
Update Go to version 1.20.2

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```


kylos101:
:wave: @atduarte @loujaybee , for your :eyes: , I'm not sure when you'll be considering the next changelog, but this is certainly a candidate to give users a heads up about.

Siddhant-K-code:
Will followup on this, after release of https://github.com/gitpod-io/workspace-images/pull/1017

PatrickRice-KSC:
Hey @Siddhant-K-code  - any update on this one? We use gitpod on an open source project that uses go 1.20, and it's really obnoxious right now to tell our contributors to manually install it. We could add a `task` that installs it, but we'd rather just use the workspace image.

axonasif:
> but we'd rather just use the workspace image.

👋 @PatrickRice-KSC, you can do exactly that through a custom workspace image. Please see:
- https://www.gitpod.io/docs/introduction/languages/go#setting-up-a-custom-dockerfile

You can change the `FROM gitpod/workspace-base:latest` line to `FROM gitpod/workspace-base:2023-03-24-22-45-37` and `ENV GO_VERSION=1.17` to `ENV GO_VERSION=1.20.2` from the guide above for latest ubuntu jammy and go versions.

kylos101:
@loujaybee @laushinka is this something IDE team can consider for the future?

chunks/lang-go/chunk.yaml
```diff
@@ -1,7 +1,7 @@
...
-  - name: "1.19.5"
+  - name: "1.19.7"
     args:
-      GO_VERSION: 1.19.5
-  - name: "1.20"
+      GO_VERSION: 1.19.7
+  - name: "1.20.2"
     args:
-      GO_VERSION: 1.20
+      GO_VERSION: 1.20.2
```
Siddhant-K-code, line 7: Recent stable version is `1.20.3` now. ([reference](https://go.dev/dl/#:~:text=e12384311403f1389d14cc1c1295bfb4e0dd5ab919403b80da429f671a223507-,go1.20.3.linux%2Damd64.tar.gz,-Archive))

gengjiawen: 1.20.4 now

kylos101:
@iQQBot :wave: , for example, if we merge this today, it would be good to share a related changelog entry, so that next week, when this lands as latest, folks understand their options. Let me know if you need anything else?

gengjiawen:
> @iQQBot 👋 , for example, if we merge this today, it would be good to share a related changelog entry, so that next week, when this lands as latest, folks understand their options. Let me know if you need anything else?

put this in https://github.com/gitpod-io/workspace-images/blob/main/CHANGELOG.md ?

kylos101:
> > @iQQBot wave , for example, if we merge this today, it would be good to share a related changelog entry, so that next week, when this lands as latest, folks understand their options. Let me know if you need anything else?
> 
> put this in https://github.com/gitpod-io/workspace-images/blob/main/CHANGELOG.md ?

@loujaybee @iQQBot @filiptronicek I'd recommend a changelog entry on the website. The users are options now are to do nothing (use latest), use the present tag that is latest now (to avoid the change), or use the tag associated with this to test the change, before it is usable as latest. Similar to [this](https://www.gitpod.io/changelog/workpace-images-os-update-and-breaking-changes) entry.

gengjiawen:
> I'd recommend a changelog entry on the website. 

tbh this will bring to much extra work for contributor, maybe put on `CHANGELOG.md` first, send weekly update to website by maintainer.

akosyakov:
I wonder why don't have the following process:
- release `latest` once a month
- nightly publish `next` for people to try and provide feedback
- merge PRs without being afraid because of above, stop merging a week or so before `latest` release

Run it all automatically as we do with VS Code, JetBrains, let the community help catch issues.

kylos101 (approved):
@filiptronicek @iQQBot I like the suggestion from @gengjiawen , about using the changelog.md file in this repo to share related chnages, it reduces friction. Approving for now, to unblock.

filiptronicek (approved):

filiptronicek:
Tested independently in a workspace

chunks/lang-go/chunk.yaml
```diff
@@ -1,7 +1,7 @@
 variants:
-  - name: "1.19.11"
+  - name: "1.19"
     args:
       GO_VERSION: 1.19.11
-  - name: "1.20.6"
+  - name: "1.20"
```
kylos101, line 5: Good call, @filiptronicek !! :tada:

PR #1050, opened 2023-03-01
Update Tailscale to 1.36.2

mrzarquon:
Refreshing tailscale to 1.36.2 ensures we now have `tailscale update` as command, allowing users to auto-update tailscale in their environment automatically in the future.


kylos101 (approved):
Thanks, @mrzarquon !

PR #1062, opened 2023-03-23
fix python bash hook for ubuntu jammy

axonasif:
## Description

The new `bash` version `5.1.16` on Ubuntu jammy possibly broke something with the `read` command we had in the python hook file. This also does some tiny optimizations and refactoring. Ubuntu focal had `bash` version `5.0.17`

## Related Issue(s)

Fixes #

## How to test

- Open https://gitpod.io/#https://github.com/axonasif/gitpod
- Check if commands from gitpod tasks are being executed

## Release Notes

```release-note
fix python bash hook for ubuntu jammy
```

PR #1063, opened 2023-03-23
fix: remove `use_pty` option for sudo

axonasif:
## Description

Having `use_pty` option enabled from the `sudoers` config has unwanted side effects for background scripts starting from `.bashrc` that calls `sudo`, it was introduced in ubuntu jammy. It doesn't seem to have any security benefits in the context of Gitpod, seems very minimal _even if it does_. For more info, see:
- https://manpages.ubuntu.com/manpages/jammy/en/man5/sudoers.5.html

It mentions:
```markdown

     use_pty           If set, and sudo is running in a terminal, the command will be run in a
                       pseudo-terminal (even if no I/O logging is being done).  If the sudo
                       process is not attached to a terminal, use_pty has no effect.

                       A malicious program run under sudo may be capable of injecting commands
                       into the user's terminal or running a background process that retains
                       access to the user's terminal device even after the main program has
                       finished executing.  By running the command in a separate pseudo-terminal,
                       this attack is no longer possible.  **This flag is off by default.**
```

## Related Issue(s)

Required for https://github.com/gitpod-io/workspace-images/pull/1062

## How to test

- Open https://gitpod.io/#https://github.com/axonasif/gitpod
- Check if `sudo` works normally by running any command
- If you want, you can check if the line containing `use_pty` option was removed: `sudo grep use_pty /etc/sudoers`, it should print nothing.

## Release Notes

```release-note
fix: remove `use_pty` option for sudo
```

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -36,6 +36,9 @@ ENV LANG=en_US.UTF-8
 ### Update and upgrade the base image ###
 RUN upgrade-packages
 
+## Remove `use_pty` option for sudo.
+RUN sed -i '/Defaults\tuse_pty/d' /etc/sudoers
+
 ### Git ###
 RUN add-apt-repository -y ppa:git-core/ppa
 RUN install-packages git git-lfs
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -36,9 +36,6 @@ ENV LANG=en_US.UTF-8
 ### Update and upgrade the base image ###
 RUN upgrade-packages
 
-## Remove `use_pty` option for sudo.
-RUN sed -i '/Defaults\tuse_pty/d' /etc/sudoers
-
 ### Git ###
 RUN add-apt-repository -y ppa:git-core/ppa
 RUN install-packages git git-lfs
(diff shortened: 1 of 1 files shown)


kylos101 (approved):
@axonasif nice find! 

We can certainly try, it's not a default with `focal`.

```
gitpod /workspace/gitpod (kylos101/into-the-fray) $ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.5 LTS
Release:        20.04
Codename:       focal
gitpod /workspace/gitpod (kylos101/into-the-fray) $ cat /etc/sudoers
cat: /etc/sudoers: Permission denied
gitpod /workspace/gitpod (kylos101/into-the-fray) $ sudo cat /etc/sudoers
#
# This file MUST be edited with the 'visudo' command as root.
#
# Please consider adding local content in /etc/sudoers.d/ instead of
# directly modifying this file.
#
# See the man page for details on how to write a sudoers file.
#
Defaults        env_reset
Defaults        mail_badpass
Defaults        secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin"
```

PR #1085, opened 2023-05-04
Update workspace image deps

kylos101:
Buildkit and Dazzle were slightl yout of date.


Furisto (approved):
Approving so that you can merge once build has completed

kylos101:
> Approving so that you can merge once build has completed

I was able to start a workspace from this branch, and `./build-combo.sh java-17`, so, we should be all set. :+1:

PR #1123, opened 2023-07-12
Use a bind mount instead of a volume

aledbf:


kylos101:
[Check out this error from the Github Action runner](https://github.com/gitpod-io/workspace-images/actions/runs/5535284229/jobs/10101363240#step:7:47)

I think it is due to [this](https://github.com/containers/skopeo/blob/a39972ca35e334b0f04ac999f57faf1ec32cdc79/vendor/github.com/containers/image/v5/pkg/docker/config/config.go#L618), which tries to write to the REGISTRY_AUTH_FILE, [after successfully logging in](https://github.com/containers/skopeo/blob/a39972ca35e334b0f04ac999f57faf1ec32cdc79/vendor/github.com/containers/common/pkg/auth/auth.go#L165)

I can recreate the error in a Gitpod workspace, but not my local machine.

kylos101 (approved):
Yep, good call.

From the GPT:
```
In your previous case, the "device or resource busy" error was likely due to how the specific application inside the container was trying to interact with the file, rather than Docker itself applying a lock. It's generally safer to mount directories rather than individual files, especially if the file needs to be written to or modified, as you found in your solution.
```

PR #1137, opened 2023-07-17
[.github] rewrite parsing of sync-containers to work with yq 3

kylos101:
🤦we're using yq 3, not yq 4, this explains why it worked fine in a Gitpod workspace, because I initially used `brew install yq` instead of `sudo pip3 install yq` to install yq.

### DIFF
--- a/.github/workflows/push-main.yml
+++ b/.github/workflows/push-main.yml
@@ -137,6 +137,7 @@ jobs:
 
       - name: 🔧 Setup copy tools
         run: |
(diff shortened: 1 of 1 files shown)


kylos101:
To test: https://gitpod.io#snapshot/1e5689e8-a2b1-4082-9945-e8e624d684d0

And run `./foo.sh`

If you want to recreate the error we were getting, do `brew install yq`, to install yq 4.

PR #1153, opened 2023-08-02
Update docker-compose

filiptronicek:
## Description

Upgrades compose to https://github.com/docker/compose/releases/tag/v2.20.2. Should also update buildx and Docker itself.

## Related Issue(s)

An attempt to address EXP-338

### DIFF
--- a/chunks/tool-docker/Dockerfile
+++ b/chunks/tool-docker/Dockerfile
@@ -13,7 +13,7 @@ RUN curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor
     && apt update \
     && install-packages docker-ce docker-ce-cli containerd.io docker-buildx-plugin
 
-RUN curl -o /usr/local/bin/docker-compose -fsSL https://github.com/docker/compose/releases/download/v2.4.1/docker-compose-linux-$(uname -m) \
+RUN curl -o /usr/local/bin/docker-compose -fsSL https://github.com/docker/compose/releases/download/v2.20.2/docker-compose-linux-$(uname -m) \
     && chmod +x /usr/local/bin/docker-compose && mkdir -p /usr/local/lib/docker/cli-plugins && \
     ln -s /usr/local/bin/docker-compose /usr/local/lib/docker/cli-plugins/docker-compose
 


.github/workflows/update-base.yml
```diff
@@ -0,0 +1,76 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set git identity
+        run: |
+          git config --global user.name $GITHUB_USER
+          git config --global user.email $GITHUB_EMAIL
+        env:
+          GITHUB_USER: roboquat
+          GITHUB_EMAIL: roboquat@gitpod.io
```
kylos101, line 25: I wonder if we need to grant `roboquat` permissions to this repo, to assume this identity. :thinking: 

I don't recall using `roboquat` for anything in `workspace-images`, yet. So the action might fail till we do something.

@securitymirco , or @aledbf , do you know what's needed for this to work? Alternatively, we could merge to main, run, and check for errors. 

Similar [action](https://github.com/gitpod-io/gitpod/blob/96bc40281b2507a5ba93c432d56dc8cb89c08ac7/.github/workflows/update-image-digest.yml).

.github/workflows/update-base.yml
```diff
@@ -0,0 +1,76 @@
+name: Update base image base
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Every Monday at midnight
+    - cron: "0 0 * * 1"
```
kylos101, line 7:
```suggestion
    - cron: "0 0 * * 4"
```

If we do on Thursday, it'll give the PR time to run checks. This way, the Engine deployer could potentially merge Friday.

Why is this timeline helpful? We'd then produce a Packer Image on Monday.

More timeline info for shipping workspace clusters here: https://www.notion.so/gitpod/Generation-eb8b655c6aa94373a73896c8055f5c9a

.github/workflows/update-base.yml
```diff
@@ -0,0 +1,76 @@
...
+        run: |
+          if [[ $(git diff --stat) != '' ]]; then
+            echo "create_pr=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@923ad837f191474af6b1721408744feb989a4c27 # v4
+        if: ${{ steps.create_pr.outputs.create_pr == 'true' }}
+        with:
+          token: ${{ secrets.ROBOQUAT_AUTOMATIC_CHANGELOG }}
```
kylos101, line 71: Looks like we need to add this secret.

.github/workflows/update-base.yml
kylos101: Looks like we need to add [this](https://github.com/gitpod-io/workspace-images/pull/1153/files#diff-3d61e19664dc79dcefdd581224a06ffd7d3615c8959c6ccb1ca7d818e855532eR71) secret?

kylos101 (commented):
Thanks for adding, @filiptronicek ! 

Added one suggestion, and pinged @securitymirco and @aledbf for help Re: integrating with roboquat.

.github/workflows/update-base.yml
securitymirco: @kylos101 I could only guess, don't know for sure 😞

filiptronicek: Thanks for catching this, @kylos101 🙏. I suppose it needs to have the proper GitHub PAT to create the PR. I will rename the secret to something better digestible and would ask @securitymirco if he could create a @roboquat PAT with the `repo` permission and add it as a workspace secret under `ROBOQUAT_PR_TOKEN`? [[docs](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#workarounds-to-trigger-further-workflow-runs)]. If you can, please also invite @roboquat to the repo so that it has access.

.github/workflows/update-base.yml
filiptronicek: Sounds good, thanks!

.github/workflows/update-base.yml
securitymirco: @filiptronicek @kylos101 I've added roboquat with write permissions on this repository. the login credentials to roboquat can be found inside the 1Password PDE vault.

laushinka:
I can approve once the build is green 👍🏽 It looks like the branch also needs to be updated with main?

laushinka (approved):

PR #1157, opened 2023-08-07
Run workflows on our runners

filiptronicek:
Building on top of ENG-489, this PR utilizes our own runners on GCP.

Fixes EXP-48


filiptronicek:
We are waiting for the proper credentials to be set, then we can test

aledbf (approved):

PR #1167, opened 2023-08-16
Delete update-base.yml

filiptronicek:
Because Renovate does the workflow's work for us, we can remove it.

cc @kylos101


kylos101 (approved):
:+1:

PR #1170, opened 2023-08-28
Reduce the number of images to sync

aledbf:


.github/promote-images.yml
```diff
@@ -4,30 +4,30 @@
 # We use a regex rule: Any tag that starts with '20'.
```
laushinka, line 4: Nit:
```suggestion
# We use a regex rule: Any tag that starts with '2023'.
```

laushinka:
Thank you again @aledbf 🙏🏽

easyCZ (approved):

PR #1171, opened 2023-08-30
[skopeo] fix skopeo version

kylos101:
## Description

quay.io only has the latest major, minor, and patch versions available

https://quay.io/repository/skopeo/stable

We were using 1.12, but, 1.13 is newer. Switching to major version of 1, to avoid future thrash when minor and patch versions change

## Related Issue(s)

Fixes n/a

## How to test

Run a job dependent on skopeo

## Documentation

/hold


kylos101:
@aledbf @laushinka can you help?

aledbf (approved):

PR #1176, opened 2023-09-04
[.github] don't skip notify on failure of dependent jobs

kylos101:
Looks like there's runner trouble, or, a need to upgrade to v4 for actions/checkout.

Ref: https://github.com/actions/checkout/issues/1448

Either way, we should notify when this job is not successful.

### DIFF
--- a/.github/workflows/dockerhub-release.yml
+++ b/.github/workflows/dockerhub-release.yml
@@ -99,7 +99,7 @@ jobs:
             docker://${{ env.DH_IMAGE_REGISTRY }}/gitpod/$IMAGE:latest
           done
   notify:
-    if: always()
+    if: always() && ((github.event_name == 'schedule') || (github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main'))
     needs: promote-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/dockerhub-release.yml
```diff
@@ -99,6 +99,7 @@ jobs:
             docker://${{ env.DH_IMAGE_REGISTRY }}/gitpod/$IMAGE:latest
           done
   notify:
+    if: always()
```
WVerlaek, line 102: this will also run and notify now even if `promote-latest`'s `if` is false (e.g. when dispatched from a branch), you could maybe check in `slack-failure` whether `promote-latest` was skipped, and then not notify?

kylos101: Good catch, thanks @WVerlaek ! I limited the `notify` job in 392a15d9b84dfc96cd252c47e8c1f3cb3faef80f to scheduled runs now (which use main), or workflow dispatch, where the context branch was main. This way, we'll still get notified on success, failure, and cancel.

WVerlaek (approved):

PR #1177, opened 2023-09-12
Create workspace image for tool-bun chunk

axonasif:
## Description

[Bun](https://bun.sh/) serves as a drop-in replacement for Node, and its binary (bun) is quite lightweight. Recently, some community members have asked if we could include it in our default image. And recently, bun hit the `v1.0` milestone, so it's probably the best time to add it, vercel also added it: https://vercel.com/changelog/bun-install-is-now-supported-with-zero-configuration

## Related Issue(s)

Fixes #

## How to test

I built the full combo and pushed to my personal dockerhub: `axonasif/workspace-full-bun`

Try here: https://gitpod.io/#https://github.com/axonasif/test/tree/bun

## Documentation

/hold

### DIFF
--- a/dazzle.yaml
+++ b/dazzle.yaml
@@ -44,7 +44,6 @@ combiner:
         - tool-brew
         - tool-nginx
         - tool-nix:2.11.0
-        - tool-bun
     - name: full-vnc
       ref:
       - full
@@ -168,6 +167,11 @@ combiner:
         - tool-nginx
         - tool-nix:2.11.0
         - tool-yugabytedb:2.19
+    - name: bun
+      ref:
+      - full
+      chunks:
+        - tool-bun
   envvars:
     - name: PATH
       action: merge-unique
--- a/dazzle.yaml
+++ b/dazzle.yaml
@@ -44,6 +44,7 @@ combiner:
         - tool-brew
         - tool-nginx
         - tool-nix:2.11.0
+        - tool-bun
     - name: full-vnc
       ref:
       - full
--- a/dazzle.yaml
+++ b/dazzle.yaml
@@ -169,7 +169,7 @@ combiner:
         - tool-yugabytedb:2.19
     - name: bun
       ref:
-      - full
+      - base
       chunks:
         - tool-bun
   envvars:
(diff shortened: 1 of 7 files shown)


aledbf:
Please don't add more content to workspace-full; instead, create a new image.

axonasif:
> Please don't add more content to workspace-full; instead, create a new image.

Sure 🙏 I'll make that change shortly.

dazzle.yaml
```diff
@@ -167,6 +167,11 @@ combiner:
         - tool-nginx
         - tool-nix:2.11.0
         - tool-yugabytedb:2.19
+    - name: bun
+      ref:
+      - full
```
filiptronicek, line 172: @axonasif would it make sense if we based this off of `base` instead in order to create a slim image like we have for Node.js? 👀

kylos101: @filiptronicek that's a good idea. If they want to base it off of `full`, I'd argue they should create and maintain a base image of their own, because it sounds like they have a specific use case (for depending on tech other than bun and what's in base).

filiptronicek (approved):
Tried the image, looks good. 

Last one thing: could you try to have a look how we could have the version as a part of the tool configuration? It would be awesome to have this auto-fixable and not have to guess when a new version comes out.

axonasif:
> Tried the image, looks good.
> 
> Last one thing: could you try to have a look how we could have the version as a part of the tool configuration? It would be awesome to have this auto-fixable and not have to guess when a new version comes out.

Thanks for pointing that out! Do you mean something like this: https://github.com/gitpod-io/workspace-images/pull/1177/commits/bea0c0c58af7cdd663bebb3d20ea5e89d9052ca2 ?

filiptronicek:
@axonasif oh nice, didn't know that supplying the version is supported, thanks a lot for taking a look. It would be wonderful if we had the version part (just 1.0.2) as BUN_VERSION in the chunk's `chunk.yaml`, like [we have for Node](https://github.com/gitpod-io/workspace-images/blob/main/chunks/lang-node/Dockerfile). I assume we won't need multiple flavors now, so we can just have one variant and call it a day.

PR #1178, opened 2023-09-15
feat: update golang

gengjiawen:


gengjiawen:
cc @kylos101

filiptronicek:
Thanks for the contribution @gengjiawen 🙏. I think it is indeed time to bump the default to Go 1.21. I will update the PR with changing the other images to Go 1.21 as well.

filiptronicek:
Looping in @kylos101 to check whether it's ok from your side to update Go to 1.21.

filiptronicek (approved):

kylos101 (approved):
Seems fine, @filiptronicek . `gitpod-dev` is already on 1.21, so no issues there. :+1: 

For anyone this ends up breaking, they could temporarily pin to an older image, until they're ready to upgrade.

PR #1179, opened 2023-09-15
Explicit specification of Rust version

filiptronicek:
Make Rust versions explicit, meaning that version setting is deterministic and autofixable. 

Also upgrades Rust to 1.72.0

### DIFF
--- a/chunks/lang-rust/Dockerfile
+++ b/chunks/lang-rust/Dockerfile
@@ -1,14 +1,17 @@
 ARG base
 FROM ${base}
 
+ARG RUST_VERSION
+
 USER gitpod
 
 # Dazzle does not rebuild a layer until one of its lines are changed. Increase this counter to rebuild this layer.
 ENV TRIGGER_REBUILD=7
+ENV RUST_VERSION=${RUST_VERSION}
 
 ENV PATH=$HOME/.cargo/bin:$PATH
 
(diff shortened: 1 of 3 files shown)


geropl (approved):
Code changes LGTM, and the image was built. :heavy_check_mark: 

Do I have to check for sth else? :thinking: 

Note that we dropped auto-updating Rust layers in the past because of the high release cadence (~6 weeks). But I think that was before dazzle, and I'm unclear how much of a problem that might be today.

filiptronicek:
Thanks for reviewing @geropl 🙏. The fact that we try to fetch the latest release is problematic not just because we don't see what the current version is, but rather that Dazzle considers it always unchanged and it lagged behind frequently. This hopefully addresses that

filiptronicek:
@kylos101 what I'm hoping for is that the [built-in autofixer](https://github.com/autofix-dev/autofix/blob/main/fixers/upgrade-rustup-toolchain.js) kicks in. Let's see 🍿

kylos101:
> @kylos101 what I'm hoping for is that the [built-in autofixer](https://github.com/autofix-dev/autofix/blob/main/fixers/upgrade-rustup-toolchain.js) kicks in. Let's see 🍿

TIL about those built-in fixers! I thought we had to add our own here: https://github.com/gitpod-io/workspace-images/tree/main/.autofix/fixers

Thanks, @filiptronicek ! Approving and merging.

PR #1201, opened 2023-10-12
Python autofix update

filiptronicek:
## Description

This changes the behavior to use https://github.com/endoflife-date/release-data/ as the source of truth for Python versions, hence removing `pyenv` as a dependency.


geropl (approved):
All checks are green. :heavy_check_mark:

PR #1202, opened 2023-10-12
Autofix GitHub Action

filiptronicek:
## Description

This PR runs Autofix inside a GitHub Action, which results in us having more control over how it runs. Specifically, the Pull Requests raised will actually result from branches inside the repository, meaning that CI will be happy and automated tests will run correctly.

## Related Issue(s)

Fixes EXP-486

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -18,7 +18,7 @@ jobs:
 
     - name: Run autofix
       run: |
-        npm -g autofix
+        npm -g autofix@0.6.0
         autofix --branches --pull-request
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -1,0 +1,22 @@
+name: Run Autofix
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/autofix.yml
```diff
@@ -0,0 +1,24 @@
...
+    - uses: actions/checkout@v4
+
+    - name: Setup Node
+      uses: actions/setup-node@v3
+      with:
+        node-version: "18.x"
+
+    - name: Run autofix
+      run: |
+        npm -g autofix
```
geropl, line 21: nit: would be cool to tie versions here somehow, so we have at least some amount of control over what code we pass our GITHUB_TOKEN to... :confused: Can we maybe hard-code a specific version of autofix at least?

But I guess we are already doing this somewhere else, and this is merely a better way to execute it, so partly out of scope here. :confused:

geropl (approved):
:heavy_check_mark: to unblock

See the comment above for a general security concern/nit

.github/workflows/autofix.yml
filiptronicek: @geropl thanks for pointing this out! I will keep Autofix updated and have now pinned the version to the latest one 🙏

PR #1206, opened 2023-10-13
lang-rust: use default profile

### LINKED ISSUE #1204: Don't use a custom profile while installing rust in workspace-rust image
### Bug description

While installing Rust, we use a custom profile called `minimal` (specified [here](https://github.com/gitpod-io/workspace-images/blob/217c2d593a42d0edb2c0546e1e7b1f94d8deac2c/chunks/lang-rust/Dockerfile#L14)). Consequently, several common components are not installed by default. As a result, the out-of-the-box experience with Rust Analyzer and its VSCode extension is not great. You have to determine which components to install in order to have a fully functional Rust Analyzer. This is not necessary if you follow the instructions at https://www.rust-lang.org/tools/install to install Rust, which is what most people do.

### Steps to reproduce

- Open https://gitpod.io/?autostart=true#https://github.com/gitpod-io/optimus/tree/891fe295fecce9c1669db3e439873880baee1113
- Then open `src/main.rs`, see that rust-analyzer is showing an error about missing `sysroot`.[[1](https://gitpod.slack.com/archives/C01KPEPLLRY/p1696889763984549?thread_ts=1696634672.510449&cid=C01KPEPLLRY)]  (because `rust-src` component isn't installed in the `minimal` profile)

### Expected behavior

It should work like a normal rust installation as https://www.rust-lang.org/tools/install .

### Example repository

https://gitpod.io/?autostart=true#https://github.com/gitpod-io/optimus/tree/891fe295fecce9c1669db3e439873880baee1113

### Anything else?

Only `--profile minimal` needs to be removed from https://github.com/gitpod-io/workspace-images/blob/217c2d593a42d0edb2c0546e1e7b1f94d8deac2c/chunks/lang-rust/Dockerfile#L14 and we're good. I believe this change will not cause any breaking change in the workspace image as this only adds (by not using the `minimal` profile, hence the default) but this may increase the image size quite a bit.

filiptronicek:
Fixes #1204

### DIFF
--- a/chunks/lang-rust/Dockerfile
+++ b/chunks/lang-rust/Dockerfile
@@ -11,7 +11,7 @@ ENV RUST_VERSION=${RUST_VERSION}
 
 ENV PATH=$HOME/.cargo/bin:$PATH
 
-RUN curl -fsSL https://sh.rustup.rs | sh -s -- -y --profile minimal --no-modify-path --default-toolchain ${RUST_VERSION} \
+RUN curl -fsSL https://sh.rustup.rs | sh -s -- -y --no-modify-path --default-toolchain ${RUST_VERSION} \
         -c rls rust-analysis rust-src rustfmt clippy \
     && for cmp in rustup cargo; do rustup completions bash "$cmp" > "$HOME/.local/share/bash-completion/completions/$cmp"; done \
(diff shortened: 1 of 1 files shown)


geropl (approved):
:heavy_check_mark:

PR #1216, opened 2023-11-11
Autofix `tool-bun`

filiptronicek:
They update a lot, so it's worth giving autofixing a shot.


filiptronicek:
It works!

image

mustard-mh (approved):
💯

PR #1289, opened 2024-02-05
Fix yugabyte-db build

filiptronicek:
Update v20 to fix build errors

### DIFF
--- a/chunks/tool-yugabytedb/chunk.yaml
+++ b/chunks/tool-yugabytedb/chunk.yaml
@@ -11,6 +11,6 @@ variants:
       PYTHON_VERSION: 3.9
   - name: "2.20"
     args:
-      YB_VERSION: 2.20.0.0
-      YB_BUILD: 76
+      YB_VERSION: 2.20.1.3
+      YB_BUILD: 3
       PYTHON_VERSION: 3.9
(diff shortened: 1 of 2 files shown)


chunks/tool-yugabytedb/chunk.yaml
```diff
@@ -11,6 +11,6 @@ variants:
       PYTHON_VERSION: 3.9
   - name: "2.20"
     args:
-      YB_VERSION: 2.20.0.0
-      YB_BUILD: 76
+      YB_VERSION: 2.20.1.3
+      YB_BUILD: 3
```
kylos101, line 15: Non-blocking:
How do you determine the value for YB_BUILD? Just curious.

kylos101 (approved):

chunks/tool-yugabytedb/chunk.yaml
filiptronicek: It's actually my first time doing this, so mayhaps I could be doing it better, but I just took what was written on their [downloads page](https://download.yugabyte.com/#linux).

kylos101: Ah, nice, I [see now](https://github.com/gitpod-io/workspace-images/blob/ef1c6b1162bb0ac8502728ea50844ecd31b83318/chunks/tool-yugabytedb/Dockerfile#L21). I missed the `b` earlier. :)

PR #1315, opened 2024-04-02
Bump Bun and Python

filiptronicek:
## Description

Bumps bun to its first minor release since v1 and patches Python.

If this is reviewed later tomorrow, I'll also push updates to Bun `v1.1.1`, it [should ship soon](https://twitter.com/jarredsumner/status/1775010398825288137).


mustard-mh (approved):
🙏

PR #1339, opened 2024-04-29
[docker] pin to v26.0.1

kylos101:
## Description

Refer to https://github.com/gitpod-io/gitpod/issues/19662 for background, this is a temporary fix.

## Related Issue(s)

Related to #19662 

## How to test

From a Gitpod workspace in this repo, test with:
```bash
./build-combo.sh base
```

## Documentation

/hold

PR #1346, opened 2024-05-07
[docker] downgrade docker-ce to 24.0.9

kylos101:
## Description

This avoids limitations with extended attributes for tar files: https://docs.docker.com/engine/release-notes/25.0/#known-limitations

The problem manifests depending on extended attributes usage with some images. For example, from within Gitpod, run:

```
docker run curlimages/curl:8.1.2 -v -k https://example.com
docker: failed to register layer: lsetxattr user.overlay.origin /bin: operation not supported.
See 'docker run --help'.
```

## Related Issue(s)

Fixes n/a
Related to https://github.com/gitpod-io/workspace-images/pull/1339

## Documentation

/hold


kylos101:
> 💯
> 
> Is there a way we can block auto-updates for this? 🤔

Yes, it looks like we can configure renovate to ignore packages: I will do in a follow-on PR: https://stackoverflow.com/a/75670202

kylos101:
> > 💯
> > Is there a way we can block auto-updates for this? 🤔
> 
> Yes, it looks like we can configure renovate to ignore packages: I will do in a follow-on PR: https://stackoverflow.com/a/75670202

I was wrong, renovate just updates base images for dockerfiles. Therefore, it'll take a human to remove this pinned version.

PR #1402, opened 2024-08-14
Update pyenv

filiptronicek:
## Description

Needed for CI checks to start passing again

### DIFF
--- a/chunks/lang-python/Dockerfile
+++ b/chunks/lang-python/Dockerfile
@@ -21,7 +21,8 @@ RUN sudo install-packages \
 	libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev \
 	# Install PYENV
 	&& git clone https://github.com/pyenv/pyenv.git ~/.pyenv \
-	&& git -C ~/.pyenv checkout 9ecd803bffaffb949fbdd8c70cb086227f6a3202 # v2.4.10 \
+	# v2.4.10
+	&& git -C ~/.pyenv checkout 9ecd803bffaffb949fbdd8c70cb086227f6a3202 \
 	&& pyenv install ${PYTHON_VERSION} \
 	&& pyenv global ${PYTHON_VERSION} \
 	&& for exec in global; do printf '%s\n' 'source "$HOME/.gp_pyenv.d/userbase.bash"' >> "$PYENV_ROOT/libexec/pyenv-$exec"; done \
--- a/chunks/lang-python/Dockerfile
+++ b/chunks/lang-python/Dockerfile
@@ -21,7 +21,7 @@ RUN sudo install-packages \
 	libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev \
 	# Install PYENV
 	&& git clone https://github.com/pyenv/pyenv.git ~/.pyenv \
-	&& git -C ~/.pyenv checkout 0963c8ac02bca616b8f91ab8b1902f48502cd813 \
+	&& git -C ~/.pyenv checkout 9ecd803bffaffb949fbdd8c70cb086227f6a3202 \
 	&& pyenv install ${PYTHON_VERSION} \
 	&& pyenv global ${PYTHON_VERSION} \
 	&& for exec in global; do printf '%s\n' 'source "$HOME/.gp_pyenv.d/userbase.bash"' >> "$PYENV_ROOT/libexec/pyenv-$exec"; done \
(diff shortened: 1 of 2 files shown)


chunks/lang-python/Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN sudo install-packages \
 	libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev \
 	# Install PYENV
 	&& git clone https://github.com/pyenv/pyenv.git ~/.pyenv \
-	&& git -C ~/.pyenv checkout 0963c8ac02bca616b8f91ab8b1902f48502cd813 \
+	&& git -C ~/.pyenv checkout 9ecd803bffaffb949fbdd8c70cb086227f6a3202 \
```
mustard-mh, line 24: How about use tags? easier to read
```
git -C ~/.pyenv checkout v2.4.10
```

filiptronicek: Yes, but we use commits to pin an exact, unchangeable version. It's a security thing.

mustard-mh: And we could make it an autofix https://github.com/pyenv/pyenv/releases as well

filiptronicek: (For instance, GitHub actions [recommend](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) to do this)

filiptronicek: I've added something human-readable in [`723edf8` (#1402)](https://github.com/gitpod-io/workspace-images/pull/1402/commits/723edf80506cb92c95c44b71fe53f16c228d19ae)

mustard-mh: > Yes, but we use commits to pin an exact, unchangeable version. It's a security thing.

Then our repo can always has security problem, i.e: we use versions like `0.0.1` to download `bun` but not check it's checksum

filiptronicek: It's true that we use tags in many places, but it's more fine with 1st parties like bun, where we get it straight from the source. For our use-cases with dazzle chunks, it also is not always possible/practical to do, so it's a balancing act.

filiptronicek: FWIW, I don't think changing this here is a good idea (having the comments would make this easier to parse for humans and pinned for machines), and also don't consider this too high stakes to invest more time into changing the workflows in this repo.

chunks/tool-yugabytedb/Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN curl -sSLo ./yugabyte.tar.gz https://downloads.yugabyte.com/releases/${YB_VE
 ENV PATH="$HOME/.pyenv/bin:$HOME/.pyenv/shims:$PATH"
 ENV PYENV_ROOT="$HOME/.pyenv"
 RUN git clone https://github.com/pyenv/pyenv.git ~/.pyenv \
-	&& git -C ~/.pyenv checkout ff93c58babd813066bf2d64d004a5cee33c0f27b \
+	&& git -C ~/.pyenv checkout ff93c58babd813066bf2d64d004a5cee33c0f27b # v2.4.10 \
```
mustard-mh, line 33: ff93c58babd813066bf2d64d004a5cee33c0f27b is https://github.com/pyenv/pyenv/commits/v2.3.10/

Forgot to update commit sha?

mustard-mh (approved):
To unblock

chunks/tool-yugabytedb/Dockerfile
filiptronicek: Yes! 🤦‍♂️
I've been playing with rebases too much haha

chunks/tool-yugabytedb/Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN curl -sSLo ./yugabyte.tar.gz https://downloads.yugabyte.com/releases/${YB_VE
 ENV PATH="$HOME/.pyenv/bin:$HOME/.pyenv/shims:$PATH"
 ENV PYENV_ROOT="$HOME/.pyenv"
 RUN git clone https://github.com/pyenv/pyenv.git ~/.pyenv \
-	&& git -C ~/.pyenv checkout ff93c58babd813066bf2d64d004a5cee33c0f27b \
+	&& git -C ~/.pyenv checkout 9ecd803bffaffb949fbdd8c70cb086227f6a3202 # v2.4.10 \
```
mustard-mh, line 33: image

Maybe `\` will be ignored

PR #1421, opened 2024-09-09
Remove dotnet 7

kylos101:
## Description

Remote dotnet 7, it is no longer in support, users should use dotnet 6 or 8, which are LTS and still in support.

This will help reduce the # of chunks and combinations we build, as well as the # of images we push to GAR and sync to Docker Hub.

Related:
* I checked, and could not find other images we can remove, such as old versions of Python, Ruby, Node, and Java.

## Related Issue(s)

Related to ENT-749

## Documentation

/hold


kylos101:
Enabled auto merge

PR #1423, opened 2024-09-10
Use placeholders to make tag name dynamic

kylos101:
## Description

For push to main, we'll only push tags for this year and month. This should help reduce how much content we sync to Docker Hub. It is presently running for too long, causing the Github Action to timeout because the private runner is only available for 4 hours.

For release to latest, we'll continue using tags for this year.

## Related Issue(s)

Fixes ENT-749

## How to test

This lands on main, and the action finishes successfully. 🤞 

## Documentation

/hold


.github/workflows/push-main.yml
```diff
@@ -76,6 +76,12 @@ jobs:
           # Any step using skopeo needs SKOPEO_SYNC_FILES env var
           SKOPEO_SYNC_FILES=$(mktemp -d)
           echo "SKOPEO_SYNC_FILES=${SKOPEO_SYNC_FILES}" >> $GITHUB_ENV
+
+          # limit what we push as time stamped images to images built in the current year and month only
+          current_year=$(date +%Y)
+          current_month=$(date +%m)
```
filiptronicek, line 82: TIL env vars are case-insensitive

.github/promote-images.yml
```diff
@@ -1,43 +1,43 @@
...
-    workspace-mysql: "2024.*"
-    workspace-mongodb: "2024.*"
-    workspace-java-11: "2024.*"
-    workspace-java-17: "2024.*"
-    workspace-java-21: "2024.*"
-    workspace-yugabytedb: "2024.*"
-    workspace-yugabytedb-preview: "2024.*"
-    workspace-gitpod-dev: "2024.*"
-    workspace-bun: "2024.*"
+    workspace-base: "CURRENT_YEAR-CURRENT_MONTH.*"
```
filiptronicek, line 7: nice!

I wonder if there are any implications on images potentially not being promoted from the end of a month? If the sync job failed to sync on the 31st, and you re-ran the following day, it would most likely not sync that release. Sounds acceptable though.

filiptronicek (approved):

.github/promote-images.yml
kylos101: I agree. It would be included "the next time" we update the base image, which happens regularly enough, I think.

.github/workflows/push-main.yml
kylos101: I can rename and make them upper case, if you like? I opted to leave as lower because we're just using them "here".

PR #1430, opened 2024-09-29
feat: upgrade to llvm 19

gengjiawen:


gengjiawen:
@filiptronicek When will this deploy to production ?

filiptronicek:
@gengjiawen it's already released in our date-tagged releases (like when you use `gitpod/workspace-full:2024-10-31-10-31-39`). It will become `latest` when we do a new manual release of the whole suite, which we expect to do in the coming weeks.

PR #1169, opened 2023-08-26
Update git-lfs to latest stable version

aledbf:
## Related Issue(s)

Fixes EXP-518

PR #1443, opened 2024-10-22
Drop Python 3.8 and introduce Python 3.13

filiptronicek:
## Description

What the title says. Python 3.8 [reached EoL](https://devguide.python.org/versions/) on October 7th and hence we don't have to push any new versions of it. On the same date, the new Python minor (3.13) was introduced.


.github/promote-images.yml
```diff
@@ -17,12 +17,11 @@
     workspace-node-18: "TIMESTAMP_TAG"
     workspace-node-20: "TIMESTAMP_TAG"
     workspace-node-22: "TIMESTAMP_TAG"
-    workspace-python: "TIMESTAMP_TAG"
```
mustard-mh, line 20: Why we need to remove `workspace-python`? (I thought images without specified versions will point to latest stable)

filiptronicek: @mustard-mh that's what we do for other images like `workspace-node` indeed, but we have not actually been updating it since [last August](https://github.com/gitpod-io/workspace-images/pull/1169), so this is just a cleanup. We can re-introduce it in the dazzle.yaml in a follow-up if we need to, but this is just syncing what we build to what we push.

mustard-mh: > @mustard-mh that's what we do for other images like workspace-node indeed

From README.md, we still have `workspace-node`, it's not removed. 

Btw, `workspace-python` is [documented](https://www.gitpod.io/docs/introduction/languages/python) in our docs as well (after this PR, we'll not maintain it anymore but users with this image configured don't know it
image

mustard-mh (approved):

PR #1537, opened 2025-03-14
Fix our use of `yq`

filiptronicek:
## Description

Looks like yq got updated under our feet somehow. This PR makes sure that doesn't scare us too much

### DIFF
--- a/.github/workflows/dockerhub-release.yml
+++ b/.github/workflows/dockerhub-release.yml
@@ -91,7 +91,7 @@ jobs:
         env:
           SKOPEO_AUTH_DIR: ${{env.SKOPEO_AUTH_DIR}}
         run: |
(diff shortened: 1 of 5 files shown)


geropl:
@filiptronicek Looks like the version bump from 3 -> 4.

Before, we would have explicit names e.g. `yq4` in some places.

If we update this:
 - How do we make sure we don't break customers?
 - How do we make sure we don't break ourselves?

IMO it would be worth investigating where exactly the bump is coming from, just to make sure we are aligned with the wider ecosystem :+1:

geropl:
>  Turns out, this was actually happening because of my dotfiles installing yq through nix, which actually meant that yq was at version 3.4.3! (and not 4.22.1, as our ws image provides).

Wow, that was actually unexpected - thanks for figuring that out! :muscle:

PR #1538, opened 2025-03-14
[lang-rust] don't install deprecated components

filiptronicek:
## Description

The `rust-analysis` and `rls` components are no longer installable. I've replaced them with `rust-analyzer` as the new LSP

See https://rust-lang.github.io/rustup/concepts/components.html#previous-components

## How to test

The build shall pass


geropl (approved):
:heavy_check_mark:

PR #1591, opened 2025-05-06
Update pyenv

filiptronicek:
## Description

Make the CI pass again (hopefully enough :P)


mustard-mh (approved):
I saw error message on main build
```
 step 01      testing "it should have a functioning java 17 installed"command=[sdk default java 17.0.14.fx-zulu && java -version && mvn -v] 
              failed                                      message="assertion failed: status == 0" result="&test.RunResult{Stdout: []byte(""), Stderr: []byte("java 17.0.14.fx-zulu is not installed on your system\n"), StatusCode: 1}" 
```

That it's failed because the java unit tests, not sure why upgrade py could fix it but it's not harmful 🚢

filiptronicek:
@mustard-mh I see, that's another thing we need to fix. The error I was seeing in builds was a missing python version, which is what this PR is fixing.

PR #1701, opened 2025-09-20
✨ Add Java 25 LTS and remove Java 23 EOL

rajadilipkolli:
## Description
Introduce Java 25 LTS version and removes Java 23 which has reached its End of life support

/hold

### DIFF
--- a/.github/promote-images.yml
+++ b/.github/promote-images.yml
@@ -38,7 +38,7 @@
     workspace-java-11: "TIMESTAMP_TAG"
     workspace-java-17: "TIMESTAMP_TAG"
     workspace-java-21: "TIMESTAMP_TAG"
-    workspace-java-23: "TIMESTAMP_TAG"
+    workspace-java-25: "TIMESTAMP_TAG"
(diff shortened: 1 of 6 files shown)


.github/promote-images.yml
```diff
@@ -38,7 +38,7 @@
     workspace-java-11: "TIMESTAMP_TAG"
     workspace-java-17: "TIMESTAMP_TAG"
     workspace-java-21: "TIMESTAMP_TAG"
-    workspace-java-23: "TIMESTAMP_TAG"
+    workspace-java-25: "TIMESTAMP_TAG"
```
kylos101, line 41: Hi @rajadilipkolli , we cannot remove workspace-java-23, some folks may still depend on it. I'll open a separate PR to introduce java 25, though.

kylos101: Hi @rajadilipkolli , the PR for Java 25 has landed, the main build takes roughly 1h, after which you should see the new image on Docker Hub, with a time stamped tag.

